PR #7, opened 2016-11-27
Add Dockerfile

gierschv:
We initially made this Dockerfile in [December 2014](https://github.com/FlatIO/faudiostream/commit/8b275c203356aa3ea05ca6e1aa915befb80ffd1b) and @mbylstra recently [asked us to make a PR](https://github.com/FlatIO/faudiostream/issues/1) here. I added some doc & updated it for the occasion.

We have the image as an [automated build in our own Docker Hub profile](https://hub.docker.com/r/flat/faust/), it may be interesting to have an automated build from this repository, so the image will always be up-to-date and ready to use.

ref: FlatIO/faudiostream#1

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,17 @@
+FROM debian:stable-slim
+MAINTAINER Vincent Giersch <vincent@flat.io>
+
+RUN mkdir /faust
+WORKDIR /faust
+COPY . /faust
+
+RUN \
+  apt-get update && \
+  apt-get install -y build-essential libssl-dev llvm libncurses5-dev libssl1.0.0 libncurses5 && \
+  rm -rf /var/lib/apt/lists/* && \
+  make && make install && \
+  make -C tools/faust2appls install && \
+  make clean && \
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM debian:stable-slim
+MAINTAINER Vincent Giersch <vincent@flat.io>
+
+RUN mkdir /faust
+WORKDIR /faust
+COPY . /faust
```
mbylstra: I was wondering if it's intentional that the faust source is copied before the apt installs, which means that the base software is re-installed whenever the faust code changes. This makes it extremely slow to test faust code changes locally, but I can see advantages: the latest debian software versions are always fetched when the source code has been updated. I'm just getting into Docker, so I was wondering if this ordering is considered best practice.

gierschv: Yes, that's a way to reduce the size of the image by 2 since we can install & remove the build dependencies in a single Docker Layer. We can do like the original version but the result of the image is much bigger (but we can take the advantage of the cache when developing).

mbylstra: I see. Thanks.

orlarey:
Done. Thanks !

PR #8, opened 2016-11-29
Fix segfault at libcode.cpp:965

jcelerier:
The "helpers" variable was NULL if writing to a string.


sletz:
Done in 7265f89c92b3f177b99f503b3d8f95e380faa467 (no yet pushed...)

jcelerier:
thanks !

PR #34, opened 2017-05-10
Add appropriate options to the SC generator

llloret:
It seems like some options that were required for SuperCollider, and that were reported had not been properly added. This PR should fix this.


llloret:
Close this PR to make an equivalent one on the master-dev branch as requested y @sletz.

PR #44, opened 2017-06-08
removed garbage and leftovers

umlaeute:
this PR removes:
- personal settings from XCode projects
- VSCode caches
- a pre-built binary for libglui.so (the sources are there and a build-system, so it should be easy to rebuild it); it was amd64/linux only anyhow...
- sphinx generated documentation in `architecture/httpdlib/html/js/svg/api` (just run `make html` in that folder)
- the ROS documentation (cherry picked from master-dev)

it also:
- adds the XCode/VStudio files we just deleted to .gitignore
- drops executable bits from source-files, images, ... (i'd rather not run a malicious png file by accident)
- removed some obvious and trivial-to-fix bashisms from some `faust2*` scripts to make them POSIX compliant


sletz:
Thanks for the PR. 

But we currently still do commit on master-dev (the test development branch), then merge on master, then merge on faust2. At some point we'll have to drop master, but still not... ((-; So your change would better be done on master-dev, so that o be prorated everywhere. 

Can you possible rework the PR this way? Thanks.

umlaeute:
ah, i see

PR #47, opened 2017-06-08
removed garbage and leftovers, cleanups, fixed the rest

umlaeute:
this is a port of #44 to the `master-dev` branch

this PR removes:
- personal settings from XCode projects
- VSCode caches
- a pre-built binary for libglui.so (the sources are there and a build-system, so it should be easy to rebuild it); it was amd64/linux only anyhow...
- sphinx generated documentation in `architecture/httpdlib/html/js/svg/api` (just run `make html` in that folder)
- the ROS documentation (cherry picked from master-dev)

it also:
- adds the XCode/VStudio files we just deleted to .gitignore
- makes the documentation buildable even if faust has not been previously installed (by falling back to the local `faust2md` script)


umlaeute:
the dropping of executable bits has been split into many small patches, mainly because my original monolithic commit on the `faust2` branch wouldn't apply on `master-dev`. so i thought it better to just split it into smaller chunks.

sletz:
OK, are they some more ? (at the end I'll just merge them with "squash" mode to have one single one commit...)

umlaeute:
@sletz i think i'm through with the executable bits :-)

about squashing: this might make it harder to actually apply them to the `faust2` branch once you think its ready...

sletz:
OK...

PR #49, opened 2017-06-13
added buttons and dynamic memory management

pingdynasty:
Brings the OWL architecture file up to date and adds support for Stéphane's latest `dsp_memory_manager`

### DIFF
--- a/architecture/owl.cpp
+++ b/architecture/owl.cpp
@@ -36,10 +36,10 @@
 #ifndef __FaustPatch_h__
 #define __FaustPatch_h__
 
-#include "StompBox.h"
 #include <cstddef>
 #include <string.h>
 #include <strings.h>
+#include "Patch.h"
 
 #ifndef __FaustCommonInfrastructure__
 #define __FaustCommonInfrastructure__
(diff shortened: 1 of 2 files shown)


sletz:
Seems like the following code in incorrect (you are destroying the fDSP object "twice"..)

 ~FaustPatch(){
 	mem.destroy(fDSP);
 	delete fDSP;
       }

If you want to use the custom allocator for the mysdsp object memory, then C++ placement new should be used, that is:

fDSP = new (mem.allocate(sizeof(mydsp)) mydsp(); in  FaustPatch() constructor

and

fDSP->~mydsp() (even if ~mydsp() does nothing but you can see the point...)
mem.destroy(fDSP);

pingdynasty:
destroy twice: correct, I'll fix it.

placement new: Since the memory allocator is using `new`/`delete` I am using the same mechanism in the patch ctor.

I'll replace this with a better PR.

sletz:
If you use fDSP = mydsp();  then the standard C++ operator new is called.. you don't control it. Using placement new with your custom allocation then you fully control the allocation/destroy steps...

sletz:
You can just update/Correct the PR with a new commit I guess...., I can possibly using "smash merge" to have ann intermediate commits in a unique big one.

sletz:
Ok now I read the comment "Simple heap based memory manager * Uses overloaded new/delete operators on OWL hardware.", so OK for the contractor part then...

pingdynasty:
I've added placement new as you suggested (separate PR). It amounts to the same thing atm, but if we change the memory manager in the future it will be useful to have the dsp object allocated in the same way.

PR #51, opened 2017-06-13
bringing code up to date with OWL architecture and script changes

pingdynasty:
Added support for buttons in architecture file.
Fixed issue with `faust2owl` script.
Added custom memory management for OWL architecture.
Added -mem flag to OWL script.

### DIFF
--- a/architecture/owl.cpp
+++ b/architecture/owl.cpp
@@ -10,12 +10,12 @@
 
 /************************************************************************
     FAUST Architecture File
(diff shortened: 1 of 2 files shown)


sletz:
So with this simple OwlMemoryManager, it means that static data (possibly allocated by "classInit") will not be deallocated for now. Is that OK for now?

pingdynasty:
It's not ideal but in our environment it works.

PR #60, opened 2017-07-05
Creation unity plugin files

theodupre:


sletz:
Could you please move the unity.cpp and unityplugin.cpp in the newest "architecture/unity" folder and update the script accordingly ?

PR #66, opened 2017-07-15
Changed the coefficient of lip filter according to original STK's way.

tomoyanonymous:
In original STK source,coefficient a1 is not squared (https://github.com/thestk/stk/blob/master/src/BiQuad.cpp & https://github.com/thestk/stk/blob/master/src/Brass.cpp).

 Or, Should it be written by using `fi.resonlp`?


rmichon:
Indeed... Thanks for catching that!. Can you do a pull request on master-dev instead of faust2? I'll push it whenever after you've done it. Thanks!

tomoyanonymous:
Thanks Michon. I did that and made new request #67.

PR #69, opened 2017-07-19
support supernova

carlocapocasa:
These definitions are required to support supernova, the  multiprocessor SuperCollider server. They have no effect outside of Supercollider build scripts with Supernova enabled.

@LFSaw asked me to take my supernova enabler for his JPverb upstream to you.

https://github.com/supercollider/sc3-plugins/pull/149

### DIFF
--- a/architecture/supercollider.cpp
+++ b/architecture/supercollider.cpp
@@ -531,4 +531,10 @@ FAUST_EXPORT void load(InterfaceTable* inTable)
 #endif // NDEBUG
 }
 
+#ifdef SUPERNOVA 
+extern "C" FAUST_EXPORT int server_type(void) { return sc_server_supernova; }
+#else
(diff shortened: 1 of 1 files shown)


vivid-synth:
I don't know if I'm reading this wrong, but don't we want to support both supernova and scsynth if ```SUPERNOVA``` is on?

carlocapocasa:
@vivid-synth Yes.

sletz:
"support both supernova and scsynth" ==> ? so : is the patch correct to do that?

carlocapocasa:
Absolutely!

What made you doubt?

carlocapocasa:
> "support both supernova and scsynth" ==> ? so : is the patch correct to do that?

Absolutely!

It is borrowed from the SuperCollider PlugLoad macro that is used by most native UGens to achieve this.

PR #70, opened 2017-07-20
also build supernova lib with faust2supercollider

carlocapocasa:
I have modified the faust2supercollider script to additionally build the supernova version on linux.


orlarey:
Supernova support should be an option (`-supernova`) as it creates additional dependencies besides Supercollider.

carlocapocasa:
Yes, I will implement this.

carlocapocasa:
Oh, already done! Thanks Yann.

PR #80, opened 2017-08-30
JSONUIDecoder.h : fix map template >>

acarabott:
Added space so template doesn't break.


sletz:
Fixed in b21340e70e072eb56b64e24fe88777200b01058d

PR #82, opened 2017-08-31
add gitignore for generated files

acarabott:
ignores generated files from httpdlib, osclib, compiler, faustremote and sound2faust


sletz:
Thanks !

PR #85, opened 2017-09-21
[coreaudio-ios-dsp.h] modifications to be compatible with bluetooth devices

RuolunWeng:
Hello Faust,

Here is the PR form my fork for 'coreaudio-ios-dsp.h', now it should be compatible with bluetooth devices like headphone and loudspeaker.

Allen


sletz:
Thanks, finally did the merge manually.

PR #92, opened 2017-10-26
Eliminate warnings/errors found by PVS-Studio analyzer

AskePit:
Hello!
I'm a member of the Pinguem.ru competition on finding errors in open source projects. A list of bugs, found using PVS-Studio analyzer:

- [V730 ](https://www.viva64.com/en/w/v730/)Not all members of a class are initialized inside the constructor. Consider inspecting: hasRealInput, hasRealOutput. schema.h 63 (fixed by point() constructor introduction)
- [V773 ](https://www.viva64.com/en/w/v773/)Visibility scope of the 'tmpList' pointer was exited without releasing the memory. A memory leak is possible. lateq.cpp 254
- [V792 ](https://www.viva64.com/en/w/v792/)The 'isNum' function located to the right of the operator '&' will be called regardless of the value of the left operand. Perhaps, it is better to use '&&'. fmodprim.cpp 42
- [V792 ](https://www.viva64.com/en/w/v792/)The 'isNum' function located to the right of the operator '&' will be called regardless of the value of the left operand. Perhaps, it is better to use '&&'. powprim.cpp 41
- [V792 ](https://www.viva64.com/en/w/v792/)The 'isNum' function located to the right of the operator '&' will be called regardless of the value of the left operand. Perhaps, it is better to use '&&'. remainderprim.cpp 36
- [V646 ](https://www.viva64.com/en/w/v646/)Consider inspecting the application's logic. It's possible that 'else' keyword is missing. text.cpp 201
- [V792 ](https://www.viva64.com/en/w/v792/)The 'isOne' function located to the right of the operator '|' will be called regardless of the value of the left operand. Perhaps, it is better to use '||'. mterm.cpp 479
- [V518 ](https://www.viva64.com/en/w/v518/)The 'malloc' function allocates strange amount of memory calculated by 'strlen(expr)'. Perhaps the correct variant is 'strlen(expr) + 1'. sourcefetcher.cpp 507
- [V518 ](https://www.viva64.com/en/w/v518/)The 'malloc' function allocates strange amount of memory calculated by 'strlen(expr)'. Perhaps the correct variant is 'strlen(expr) + 1'. sourcefetcher.cpp 538
- [V518](https://www.viva64.com/en/w/v518/) The 'malloc' function allocates strange amount of memory calculated by 'strlen(expr)'. Perhaps the correct variant is 'strlen(expr) + 1'. sourcefetcher.cpp 601
- [V773 ](https://www.viva64.com/en/w/v773/)The function was exited without releasing the 'f' handle. A resource leak is possible. enrobage.cpp 310
- [V669 ](https://www.viva64.com/en/w/v669/)The 'pt1' argument is a non-constant reference. The analyzer is unable to determine the position at which this argument is being modified. It is possible that the function contains an error. doc_lang.cpp 164
- [V1003 ](https://www.viva64.com/en/w/v1003/)The macro 'DEFAULT_PAGE_BUF_SIZE' is a dangerous expression. The expression must be surrounded by parentheses. sourcefetcher.hh 50

Not fixed but take a look at it:
- [V523 ](https://www.viva64.com/en/w/v523/)The 'then' statement is equivalent to the 'else' statement. doc_compile.cpp 1221
- [V501 ](https://www.viva64.com/en/w/v501/)There are identical sub-expressions to the left and to the right of the '/' operator: (x * p) / (x * p) compatibility.cpp 190

There are actually more warnings were found but they are not so interesting.

### DIFF
--- a/compiler/documentator/doc_lang.cpp
+++ b/compiler/documentator/doc_lang.cpp
@@ -54,7 +54,7 @@ static const string			gDocTextsDefaultFile = "mathdoctexts-default.txt";
 
 static void			importDocStrings(const string& filename);
 static void			getKey(string& s, string& key, size_t& pt1);
-static void			getText(string& s, size_t& pt1, string& text);
+static void			getText(const string& s, size_t pt1, string& text);
 static void			storePair(const string& key, const string& text);
 
 static void			printStringMapContent(map<string,string>& map, const string& name);
--- a/compiler/documentator/doc_lang.cpp
+++ b/compiler/documentator/doc_lang.cpp
@@ -161,7 +161,7 @@ static void getKey(string& s, string& key, size_t& pt1)
 }
 
 
-static void getText(string& s, size_t& pt1, string& text)
+static void getText(const string& s, size_t pt1, string& text)
 {
 	/* Capture the text on the current line. */
 	size_t pt2;
--- a/compiler/documentator/lateq.cpp
+++ b/compiler/documentator/lateq.cpp
@@ -248,9 +248,9 @@ vector<list<string> > Lateq::makeUISignamesVector(const multimap<string,string>&
 		} else {
 			++vIndex;
 			uiTypesMap.insert(pair<char,unsigned int>(type, vIndex));
-			list<string>* tmpList = new(list<string>);
-			tmpList->push_back(signame);
-			uiSignamesVector.push_back(*tmpList);
+			list<string> tmpList;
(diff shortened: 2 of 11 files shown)


sletz:
Thanks. 

That would be better if you prepare the PR for the faust2 branch, which is going to be the main branch soon.

sletz:
Actually doing that manually in faust2. Thanks for the initial PR. Will be committed soon.

PR #97, opened 2017-11-23
Minimum dataspace array sizes of 1 for Microsoft C++ compiler.

gogins:
The Microsoft compiler does not permit the declaration of zero size arrays. This commit makes each array in the dataspace struct for the Csound template have a minimum size of 1.


sletz:
Merged thanks.

PR #99, opened 2017-11-24
The original solution pushed the dataspace elements off by 1 element for inputs, outputs, and actives. 

gogins:
Csound strictly expects outputs and inputs to be in index order in the opcode structure. Adding 1 to sizes pushes indexing off by 1 for outputs, inputs, and controls.

With this commit, if any of FAUST_OUTPUTS, FAUST_INPUTS, or FAUST ACTIVES is generated with a size of 0, then aout, ain, or ktl is not declared in its original location at all. Instead, it is moved to the end of the opcode structure and declared with a size of 1. The compiler will find definitions for these symbols so that compilation will succeed, but the generated DSP code should not actually reference these objects.


sletz:
Merge --squash done.

Issue #100, opened 2017-11-24
Please implement default values of controls in generated Csound opcodes


gogins:
Please implement default values of controls in generated Csound opcodes. This may be advisable for other architectures as well.

When the user has a working prototype in FaustLive or some other development environment, then the current values of user interface controls should be saved as the default/initial values of the DSP control variables.

In Csound, it is possible to declare optional k-rate parameters with a default value of -1 as "J". Then in the generated interface code, the values of the incoming parameters would be tested. If they were -1, the incoming parameter would automatically be set to the default value from the Faust DSP code. If they were not -1, the incoming parameter would be used with its current value.

In cases where the generated opcode used many parameters, this could save the user of the opcode from having to specify the value of every parameter in the Csound orchestra code.

There might be other ways to implement this...

sletz:
Default values of all controllers are given when calling ` p->DSP->buildUserInterface(p->interface);` (line 199 of csound.cpp). The used CSUI class has a set of `addCheckButton`, `addVerticalSlider` that receive the `FAUSTFLOAT init,` parameter. Could the values be used at this place ?

sletz:
Note that there is also the (dynamic libfaust + LLVM) Faust/csound link that should possibly be checked also : see cashttps://github.com/grame-cncm/faust/tree/master-dev/embedded/faustcsound

gogins:
I can't look at that until I have LLVM built to work with Faust. But I will
keep it in mind.

Regards,
Mike


-----------------------------------------------------
Michael Gogins
Irreducible Productions
http://michaelgogins.tumblr.com
Michael dot Gogins at gmail dot com

PR #102, opened 2017-11-25
Enable default parameter values in DSP code to function as default Csound opcode parameter values (#100).

gogins:


sletz:
Merge but the code was incorrect , correct this way OK?

```
void copyfrom(MYFLT* mem[]) {
         for (unsigned int i=0; i<vZone.size(); i++) {
-            if (*(mem[i] != FL(-1.)) {
+            if (*(mem[i]) != FL(-1.)) {
                 *vZone[i] = *(mem[i]);
             }
         }
```

gogins:
Yes, thanks.

Mike

PR #108, opened 2017-12-20
Change in makefiles: make -> $(MAKE)

yurivict:


agraef:
@yurivict looks ok to me, but next time please take 2 min and write a short description so that everybody understands what problem you're trying to solve (or link to a related issue reported earlier which discusses the problem). That's just a matter of courtesy -- it's the duty of the reporter to explain his or her problem, not upstream's. If the description is blank, a developer might not even bother with the actual content. We're all terribly busy all the time, just like you, so that might have been the reason that this apparently got ignored.

@sletz I see why this is needed on the BSDs at least (besides, it's recommended practice: https://www.gnu.org/software/make/manual/html_node/MAKE-Variable.html), I tested it with latest from master-dev on both Linux (Arch) and macOS (Sierra), I'd say it's good to go in. Even if you're working on an alternative cmake build system right now, I hope that our good old Makefiles don't get scrapped right away, and then we'd like them to work, even on BSD. If you don't have any complaints, I'll go ahead and merge. Ok?

sletz:
Yann told (21 demure 2017) :` Unfortunately, $(MAKE) is not predefined in some systems we support (for instance macOS). `, so are we sure now?

sletz:
So after discussion with Dominique, OK for the merge.

yurivict:
This MacOS makefile seems to use ```$(MAKE)```:
https://sourceforge.net/p/passwordsafe/git-code/ci/d61435b93a245be03e89f849b0a56a3a7e166338/tree/Makefile.macos

agraef:
@sletz:
> Yann commented (21 decembre 2017) :Unfortunately, $(MAKE) is not predefined in some systems we support (for instance macOS)., so are we sure now?

Sure is. It's GNU make, which had this for ages if memory serves me.

> So after discussion with Dominique, OK for the merge.

Ok, in it goes...

orlarey:
I was clearly wrong! Sorry for that

2018-02-06 2:45 GMT-08:00 Albert Graef <notifications@github.com>:

PR #114, opened 2018-01-12
fix bitcode compile for pure-lang/pd-faust

boqs:
this change required in order to successfully run "make bitcode" for pure-lang's pd-faust framework

(@agraef - still haven't managed to get pd-faust compiling/working correctly against recent faust, pd & pure-lang repositories but this is surely a step in the right direction)


agraef:
@boqs, cool, thanks. I just pulled this over upstream.

Issue #109, opened 2017-12-24
Version 2-5-10 doesn't install stdfaust.lib


yurivict:
Version 2.1.0 did install it, and the test program from http://faust.grame.fr/examples/2015/09/30/oscillator.html worked.

Version 2.5.10 doesn't install it, and this test program fails:
```
$ faust2jaqt sine.dsp
ERROR : /usr/local/bin/faust2jaqt cannot find Faust library dir (usually /usr/local/share/faust)
ERROR : unable to open file stdfaust.lib
```

Please also note the first error: it's bogus because the directory ```/usr/local/share/faust``` exists.

pingdynasty:
I'm getting the same thing, `stdfaust.lib` is not installed or even built. Building with current faust master, Version 2.5.13, on Ubuntu 16.04 64bit.
Version 2.5.16 on master-dev branch builds and installs fine.

agraef:
The library is now in a git submodule. You'll have to clone the git source (`git clone https://github.com/grame-cncm/faust.git`), then just running `make` will do the necessary `git submodule update --init`. This doesn't work with the tarball since it's not a git repo.

agraef:
@sletz, maybe we need to provide a proper self-contained tarball for people who are not familiar with git submodules. I already have that in my `$(debsrc)` target in the main Makefile, we'd just have to massage that a bit so that you could do a `make dist` and upload the resulting tarball whenever you tag a release. Should I do a PR so that you can have a look?

sletz:
"Should I do a PR so that you can have a look?" yes please ! (thanks)

agraef:
Will do. (Might take a day or two, though, as I'm currently busy preparing courses for tomorrow and Friday.)

pingdynasty:
I see - I'm using a git repo. Installation is fine now if I first do a `git submodule update` manually.

 It looks to me like the `git submodule update --init` will only happen if I `make debsrc`.

agraef:
@pingdynasty Only certain targets do this, but IIRC `make` is one of them (in the latest HEAD at least). In any case, it's a good idea to just run it manually after a `git pull` before you do a `make` or `make world` so that you can be sure.

agraef:
@sletz there you go: PR #116. This also includes a new dist-snapshot target which uses the full Debian package version (useful to create git snapshots; contents is the same).

[...]

PR #116, opened 2018-01-19
Fix up the dist target so that it produces a self-contained tarball w…

agraef:
…hich includes all submodules. Also added a secondary dist-snapshot target which includes the git revision and hash in the version number.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -277,9 +277,25 @@ uninstall :
 	rm -f $(prefix)/bin/faustbench
 	rm -f $(prefix)/share/man/man1/faust.1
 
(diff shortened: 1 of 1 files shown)


agraef:
This fixes issue #109.

sletz:
Merged thanks.

agraef:
Ok, so how about tagging 2.5.17 (**and** uploading the tarball after running `make dist`) so that you can test it for yourself? Never mind, I'll keep bugging you with this until you finally give in and make that release so that I can hand my Arch faust package over to David Runge at last. 😄

PR #128, opened 2018-01-31
0 9 67 mr2

florian-grond:
For the better integration of Faust generated CPP files of SuperCollider Ugens as SC3plugins we would like to suppress the debug messages at server start and Ugen launch.


sletz:
What Faust branch are you starting from? This patch seems to use an old mr (= multi rate) branch. 
So I don't understand why this patch touch also those files, this is clearly not correct. Your patch is supposed to touch architecture/supercollider.cpp and possibly tools/faust2appls/faust2supercollider  script.

You should probably start from a fresh clone (one master-dev) and start working from that point. Thanks!

florian-grond:
Sorry for that, was an old branch, will do as you suggest!



www.grond.at

PR #130, opened 2018-02-02
replacing NDEBUG macro with the more specific F2SC_DEBUG_MES...

nicobou:
… in the supercollider cpp file and in faust2supercollider script.

This PR is motivated by the need for inclusion off faust generated cpp files into the official sc3-plugins repository, which is often compiled in debug mode. Without this patch, the Faust prints are distracting the supercollider console, while not debugging the faust related development. This could pave the way to seamless integration of faust generated UGens.

The change proposed here makes the `Print` invocation in the cpp file very specific and controllable through the `-dm` option in the faut2supercollider script.

### DIFF
--- a/architecture/supercollider.cpp
+++ b/architecture/supercollider.cpp
@@ -437,7 +437,7 @@ void Faust_Ctor(Faust* unit)  // module constructor
                 }
                 SETCALC(Faust_next_copy);
             }
-    #if !defined(NDEBUG)
+    #if defined(F2SC_DEBUG_MES)
             Print("Faust[%s]:\n", g_unitName);
             Print("    Inputs:   %d\n"
                   "    Outputs:  %d\n"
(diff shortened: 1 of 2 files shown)


sletz:
Seems fie, AFAICA, @fgrond is is ok for you?

florian-grond:
Yes @sletz, @nicobou  and I did it together, will be great to have this patch. We are eager to make Faust and SuperCollider work together as closely as possible. There is so much great Faust stuff out there waiting to make it into the official SC3plugins repo. Thanks!

sletz:
Merge done, thanks!

sletz:
@fgrond, @nicobou : what would also great would be to add Faust dynamic compilation (using libfaust + LLVM) into an SCPlugin. This has been discussed several times on the SC list, but nobody never really started working on that. I could help to make in happen (helping on the Faust integration side), if a SC developer does the hard work.

nicobou:
Thank you @sletz ! Thank you also for pointing us to the Faust dynamic compilation, I didn't know about this before.

sletz:
dynamic Faust compilation: here http://faust.grame.fr/news/2016/01/12/using-dynamic-compiler.html, past discussions on SC list, various GRAME papers, example code here : https://github.com/grame-cncm/faust/blob/master-dev/tools/benchmark/dynamic-jack-gtk.cpp

Feel free to ask more if you start to work on that.

florian-grond:
Thanks a lot @sletz  for merging this. Dynamic compilation is definitely interesting thanks for passing on the pointers.

Issue #131, opened 2018-02-04
Faust prefix in SuperCollider Ugens


florian-grond:
@sletz  just merged a much-appreciated patch which helps to better integrate Faust generated Ugens in SuperCollider.
The last obstacle, it seems, to comply with the SuperCollider Ugen Naming convention is the Faust prefix which is automatically added to each Faust generated Ugen.

This prefix shows up in two places:
1) the C++ source code where it can be removed during compilation. So when compiling Faust generated SC3 plugins one needs to set_property(TARGET ... APPEND PROPERTY COMPILE_DEFINITIONS SC_FAUST_PREFIX="\"\"")

2) it also shows up in the SuperCollider language class file: e.g. generated with faust2sc
**Faust**HOAEncoder1 : MultiOutUGen
....
 name { ^"**Faust**HOAEncoder1" }

I would be great to have a flag in **faust2supercollider**, which - when keeping the C++ sources with the -ks flag - allows controlling the SC_FAUST_PREFIX in the C++ sources setting it to " " and generating the SuperCollider language file without the Faust prefix too.

This would allow adding large collections of Faust libraries to the SC3plugins, without the need to check for each Faust generated Ugen the prefix having to set the COMPILE_DEFINITIONS and editing the SuperCollider class file.

@brianlheim, @patrickdupuis, @sletz, @telephon, @LFSaw, @jamshark70  
For Faust to leave traces in the Ugens I suggest adding an **info** method to the SuperCollider class file which prints to the SuperCollider post window:
This info method could contain info about the Faust version with which it was generated, the author of the Faust code and the license, essentially everything that is declared in the Faust file under:
declare name       
declare version     
declare author     
declare license     
declare copyright

sletz:
"I would be great to have a flag in faust2supercollider" => can you @lorian-grond and/or @nicobou prepare a PR for that? Thanks.

florian-grond:
Will try if @nicobou wants to assist, at the earliest next Friday.
@brianlheim, @patrickdupuis, @telephon, @LFSaw, @jamshark70
any opinion about the info method in the SuperCollider class file?

nicobou:
@florian-grond sure, I want to assist.

florian-grond:
yes, I was also thinking to post on request.



www.grond.at

brianlheim:
[...]

PR #132, opened 2018-02-05
adding the -noprefix option to faust2supercollider script

nicobou:
We added the option -noprefix to the faust2supercollider script. Default behavior remains the same as previous: "Faust" is used as prefix. 

With the -noprefix option, the resulting UGen class name comply with supercollider3 naming conventions. 

This follows the discussion #131


florian-grond:
great thanks! the SC community will be very happy :-)

PR #136, opened 2018-02-11
Revert rev. 6ec9f8f6 which trips a compiler bug (cf. #127)

agraef:
Revert "Fix 2017-12-16 bug by doing a deeper evaluation of expressions involved in pattern matching"

This reverts commit 6ec9f8f62e3da18fc950cd16734b03ddc4e4d2df.
Fixes issue #127.

### DIFF
--- a/compiler/evaluate/eval.cpp
+++ b/compiler/evaluate/eval.cpp
@@ -151,13 +151,13 @@ static Tree real_a2sb(Tree exp)
 			return result;
 
 		} else if (isBoxAbstr(abstr, var, body)) {
-			// Here we have remaining abstraction that we will try to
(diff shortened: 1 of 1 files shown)


orlarey:
@agraef, Albert, this commit was needed for other files to compile correctly. Therefore before reverting anything we must first understand what's going on...

agraef:
Ok, I understand. Fortunately, this doesn't affect any of my own programs, and I haven't really looked at @magnetophone's example. But the error message reported in #127 looks weird indeed, it looks like the evaluator didn't do the right thing there.

orlarey:
Yes probably, I'm just taking a look at it...

2018-02-10 16:43 GMT-08:00 Albert Graef <notifications@github.com>:

PR #139, opened 2018-02-14
Add regression test for #127.

agraef:
As discussed in #127. Anyone who's running the regression tests at Grame please review and merge, thanks.


agraef:
Note that @magnetophon edited this two times in aforementioned issue since I submitted this, but I checked this (his original) version ot the test, so there you go.

magnetophon:
Thanks!

PR #155, opened 2018-03-16
fix runtime output directory

dfober:


agraef:
Hi @dfober, I think that something has gone awry with your fork from which you are doing these pull requests. It's not normal that each of your PRs lists 180+ commits for each trivial change. That makes it basically impossible to follow your PRs. Maybe you need to rebase your master-dev branch on upstream? If that doesn't help then maybe your fork needs to be recreated from scratch.

dfober:
I agree, wasn’t aware of any solution. I’ll try the ‘rebase’ one.
Thanks for the tip.

agraef:
Hi Dominique, the best workflow that I could figure out is that, if you're running your own fork of a git repo, then you keep master (or, in our case, master-dev) 100% in sync with upstream *at all times* and do your own development on a secondary branch -- or many of them, if you're using feature branches, as I do. The main advantage of feature branches is that they let you work on as many different tasks as you like, simultaneously, and you can also rebase and squash them as needed, e.g., if someone like Stephane asks you to tidy up your pull request. :)

You then just submit the PR from the feature branch as usual. When the PR gets merged, you can simply delete the feature branch and update your master-dev branch by merging upstream/master-dev again.

Basically, you can and should do the latter any time when upstream/master-dev pulls ahead of your master-dev branch. One thing that is very important when using this workflow: Never **ever** let your master-dev diverge from upstream/master-dev, or the house will be on fire. ;-) That's probably what happened in your case.

Technically, the way to do this is to simply add upstream as a remote to your working copy of your repo, e.g.:

    git remote add upstream git@github.com:grame-cncm/faust.git

You only need to do that once. Afterwards, to merge with upstream:

    git checkout master-dev
    git fetch upstream
    git merge upstream/master-dev
    git push

And to rebase your feature branch on current master-dev you can then simply do:

    git checkout my-feature-branch
    git rebase master-dev
    git push -f

You can do the force-push safely while a PR from `my-feature-branch` is already in progress. Just **never never never** rewrite history on the master/master-dev branches, the master branches ought to maintain sane linear histories under all but the most exceptional circumstances.

Cf. https://help.github.com/articles/configuring-a-remote-for-a-fork/ and https://help.github.com/articles/syncing-a-fork/

Note that all that juggling with a separate upstream remote isn't needed if you just work on feature branches directly in the mother repo. Makes things a little easier.

(Maybe some kind soul could add these instructions to the wiki?)

PR #166, opened 2018-03-26
LLVM 3.4 support

agraef:
Add back support for LLVM 3.4 where `llvm-config` doesn't have `--system-libs`.

Yes, that's a really old version, but still required to build some LLVM applications (including Pure) in MacPorts, because reportedly it works better than LLVM 3.5 there. And adding back support for this version is trivial, it's just a minor configuration issue.


agraef:
I understand that LLVM 3.4 is not going to be supported in the LLVM backend any more. But with this change you can at least do a standard build without libfaust and have it work also with LLVM 3.4.

PR #182, opened 2018-05-18
Fixed C API - exceptions management

pierreguillot:
The functions `createDSPFactoryFromFile()` and `createDSPFactoryFromString` can generate exceptions so I think it would be great to catch them and to return an error (for the moment the program crashes because we can't manage the C++ exceptions in C).


sletz:
Good idea, thanks.

PR #184, opened 2018-05-21
extract file name in faust2png

astellon:
remove the extension from the command line argument to match the directory name created by faust2svg


sletz:
Can you give a description?

sletz:
@orlarey OK for you?

PR #187, opened 2018-05-29
Add a CMake option for the runtime executable target

pierreguillot:
The **faust** runtime executable is not always necessary and can be a bit annoying when using FAUST as a subdirectory in another CMake project (using `add_subdirectory()`). This option allows to remove the target **faust** (runtime executable  target) if needed - if a project only requires the static and/or the dynamic libraries for example. The target **faust** is always enable by default.


sletz:
Thanks. I'll let @dfober have a look...

dfober:
done ! thanks !

PR #215, opened 2018-09-10
Rust portaudio backend

programLyrique:
The rust backend could use jack as architecture.
I have added support to portaudio, through rust-portaudio.
I have put a faust2rportaudiorust in the tools directory.

rust-portaudio does not support non-interleaved streams so I had to use some unsafe functions to transform a pseudo interleaved stream into non interleaved streams, without copying...


sletz:
Thanks. Could you also modify the tools/faust2appls/Makefile file so that the new `faust2portaudiorust` script is properly installed and deinstalled ?

PR #219, opened 2018-09-17
using eval() in strict mode / ES6

Fr0stbyteR:
These evals are not working in some ES6 environments like Babel/Webpack, functions defined by eval disappear in the scope.
Like in this post [https://github.com/babel/babel/issues/4425](https://github.com/babel/babel/issues/4425)
Still not understanding why the compiler send this JSON with a function, not directly the stringified JSON.

### DIFF
--- a/architecture/webaudio/webaudio-wasm-wrapper.js
+++ b/architecture/webaudio/webaudio-wasm-wrapper.js
@@ -643,17 +643,20 @@ faust.readDSPFactoryFromMachineAux = function (factory_name1,
         factory.module = module;
 
         // 'libfaust.js' wasm backend generates UI methods, then we compile the code
-        eval(helpers_code1);
-        factory.getJSON = eval("getJSON" + factory_name1);
-        factory.getBase64Code = eval("getBase64Code" + factory_name1);
+        //eval(helpers_code1);
+        //factory.getJSON = eval("getJSON" + factory_name1);
+        //factory.getBase64Code = eval("getBase64Code" + factory_name1);
 
         try {
-            factory.json_object = JSON.parse(factory.getJSON());
+            factory.json_object = JSON.parse(helpers_code1.match(/getJSON.+?return\s"(\{.+?)";}function/)[1].replace(/\\/g, ""));
+            factory.base64Code = helpers_code1.match(/([A-Za-z0-9+/=]+)";\s\}/)[1];
         } catch (e) {
             faust.error_msg = "Error in JSON.parse: " + e;
             callback(null);
             throw true;
         }
+        factory.getJSON = () => { return factory.json_object; };
+        factory.getBase64Code = () => { return factory.base64Code; }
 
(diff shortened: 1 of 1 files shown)


sletz:
Your patched webaudio-wasm-wrapper.js file does not work anymore if I deploy it on the test pages here :  https://faust.grame.fr/dynamic/faustlive-wasm.html.

What is your use-case basically?

Fr0stbyteR:
> Your patched webaudio-wasm-wrapper.js file does not work anymore if I deploy it on the test page here : https://faust.grame.fr/dynamic/faustlive-wasm.html.
> 
> What is your use-case basically?

I'm using Faust-Wasm in a webaudio environment created with Babel/Webpack
[https://fr0stbyter.github.io/webaudio-patcher/](https://fr0stbyter.github.io/webaudio-patcher/)

I saw the Issue [https://github.com/grame-cncm/faust/issues/211](https://github.com/grame-cncm/faust/issues/211) reported the same problem that eval is not working properly if faust-webaudio-wrapper is called with strict mode environment.

For me the modification work fine in my program. It's just retrieving the json and base64code with regexp.

Fr0stbyteR:
> AFAICS #211 was more a "old broken asm.js version" issue of some of our tools. We need to remove them since WASM is available now.
> I'll not merged your patch for now. I'm in the process of restructuring the JS/WASM glue code, and probably completely removing the need for "eval".
> BTW nice https://fr0stbyter.github.io/webaudio-patcher/ project ! How far is it supposed to go?

No problem, 
PS It will be nice if I can have a callback to know when factory is ready. I did it by my side, not pulling request cuz it changes the initialization, by using `FaustModule({postRun : callback});`.

The aim of my project is to build an environment like Max / OpenMusic but in web, so that it can be easier to build or play music with others.

sletz:
" callback to know when factory is ready." this is already the case: `faust.createDSPFactory = function (code, argv, callback)` takes a callback which is called with the factory when ready

Fr0stbyteR:
> " callback to know when factory is ready." this is already the case: `faust.createDSPFactory = function (code, argv, callback)` takes a callback which is called with the factory when ready

Oh I mean when FaustModule is ready (when it compiles libfaust-wasm.wasm), cuz otherwise I get an error when calling `faust.createDSPFactory()`

sletz:
Ok. You'll have to setup the callback in `faust_module['onRuntimeInitialized'] `. This will be called automatically by the Emscripten generated runtime when ready. See the following file as an example:
https://github.com/grame-cncm/fausteditorweb/blob/master/scripts/faustlive.js

Fr0stbyteR:
> Ok. You'll have to setup the callback in `faust_module['onRuntimeInitialized']`. This will be called automatically by the Emscripten generated runtime when ready. See the following file as an example:
> https://github.com/grame-cncm/fausteditorweb/blob/master/scripts/faustlive.js

Good to know that. Thx !

PR #223, opened 2018-09-29
Fix ir2dsp

jkbd:
The generated output of `ir2dsp.py` did not compile with faust due to missing environment prefixes. Also the argument order of `pm.modalModel(...)` was in wrong order.

Please note that in https://github.com/grame-cncm/faustlibraries/blob/master/physmodels.lib the environment prefix to `take` is still missing and the use of `modalFilter` is wrong. `gain(i)*modeFilter(freqs(i),res(i))` should be changed to `modeFilter(freqs(i),res(i),gain(i))`.


sletz:
@rmichon : seems correct to me, could we merge?

rmichon:
@sletz Yes! @jkbd Thanks for catching all these errors. I'll let the author of this script know that these changes were made.

PR #225, opened 2018-09-30
build fixes on archlinux

jcelerier:
I had to perform these changes to build the latest commit - the first (order of extern "C" and export) should not break anything - and I guess the second one neither since <errno.h> is included, thus making that errno declaration redundant.


jcelerier:
thanks !

PR #226, opened 2018-09-30
Add LLVM version 6.0.1 to Makefile.unix

mil:
Add LLVM version 6.0.1 to Makefile.unix


sletz:
Applied but note that your should now better use the cmake based compilation model.

PR #242, opened 2018-11-18
Unity utilities script import dll correctly on linux

### LINKED ISSUE #240: Wrong dll name in FaustUtilities Unity script on linux
A binary on linux with the name `libFaustPlugin_whatever.so` should be loaded into Unity with `DllImport("FaustPlugin_whatever")`. At the moment the automatically generated `FaustUtilities_whatever.cs` script loads this binary with `DllImport("libFaustPlugin_whatever")` when runs in the editor. Here's the source of the issue:

https://github.com/grame-cncm/faust/blob/6d00c06c39eb41312ea216bf1d0a50fc70c74045/architecture/unity/FaustUtilities_template.cs#L247-L255

`PLUGNAME` is replaced by `libFaustPlugin_whatever` while, for linux, it should be replaced by `FaustPlugin_whatever`.

A simple workaround for now is to just edit the file manually. Then it works.

BTW, I'm making my first steps with faust today and so far it's a beautiful piece of software. Thanks for that!

Nagasaki45:
Tested and works fine for both editor and standalone. Fix #240

### DIFF
--- a/architecture/unity/FaustPolyUtilities_template.cs
+++ b/architecture/unity/FaustPolyUtilities_template.cs
@@ -245,9 +245,9 @@ namespace FaustUtilities_MODEL {
 
 		private IntPtr _context;
 
-        #if UNITY_STANDALONE_OSX || UNITY_STANDALONE_WIN || UNITY_EDITOR  || UNITY_WSA || UNITY_WSA_10_0 || UNITY_IOS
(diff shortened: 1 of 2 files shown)


sletz:
Thanks !

Nagasaki45:
Thank **you** for the help and the super fast responses!

PR #285, opened 2019-02-25
Fix DLL name on iOS in the unity template

asus4:
Hi, thanks for the great repository.

I tried `faust2unity` converter and it throws DllNotFound error on iOS. the DLL name should be `__Intanal` on iOS-unity platform. 

Tested on 
- Unity Version: 2018.3.4f1
- Xcode: 10.1


sletz:
Thanks !

PR #288, opened 2019-02-27
Add OSC option in faust2android

RuolunWeng:
This is the beta test for "faust2android -osc"


sletz:
Still not tested sorry.... We ned to buy a new Android phone to actually test ((-;

PR #299, opened 2019-03-29
Report the metadata in XML output

jpcima:
This is a lack of the XML report in comparison to the JSON one.
I add the metadata from the global set, actives and passives.
It's in the form `<meta key="foo">bar</meta>`.

Remark: I have also noted that JSON lacks some information that XML has: `varname`


sletz:
Thanks !

PR #305, opened 2019-04-09
libfaust: cmake "SELF_CONTAINED_LIBRARY" to prevent library from sear…

kmatheussen:
…ching system for architecture files


sletz:
What is this PR goal?

kmatheussen:
When distributing a self-contained libfaust which includes all necessary architecture files and libraries, setting SELF_CONTAINED_LIBRARY to "on" prevents other files on the computer to be used instead.

It also prevents unnecessary disk access and error messages when compiling and creating aux files. (lots of "which: no faust in ..." error messages in the terminal)

sletz:
I don't remember the details, but seems yes. Then just try...

PR #331, opened 2019-06-11
webaudio-workletnode-standalone-wrapper: ecmascript 6 export of factory class 

lonce:
class mydsp now exported in ecmasript 6 compliant way: 

// Factory class
export default class mydsp {

so that the factory can be imported into other js modules using the import statement.
(windows name space and npm exports unchanged at the end of the file).

### DIFF
--- a/architecture/webaudio/webaudio-workletnode-standalone-wrapper.js
+++ b/architecture/webaudio/webaudio-workletnode-standalone-wrapper.js
@@ -344,7 +344,7 @@ class mydspNode extends AudioWorkletNode {
 }
 
 // Factory class
-class mydsp {
+export default class mydsp {
 
     /**
      * Factory constructor.


sletz:
The same code webaudio-workletnode-standalone-wrapper.js is also used in  `faust2webaudiowasm`  which creates standalone HTML pages.
 I tried the` export default ` but get an `Uncaught SyntaxError: Unexpected token export` when running the HTML page in Chrome Canary. 
Any idea ?

sletz:
Actually the code currently contains: 

```
// WAP factory or npm package module
if (typeof module === "undefined") {
    window.mydsp = mydsp;
    window.Faustmydsp = mydsp;
} else {
    module.exports = { mydsp };
}
```
Isn't this already enough for you needs?

lonce:
Yes - the importer would have to be a ES6 module.
<script type="module">
import foo from "./mydsp.js";
</script>

Pull request withdrawn.

PR #333, opened 2019-06-20
Fix architecture file for faust2alsaconsole

corajr:
alsa-console.cpp refers to JSONUI (see e.g. https://github.com/grame-cncm/faust/blob/f1985b49b74ec520a076e9d7f76793a5e3bea72f/architecture/alsa-console.cpp#L106) but JSONUI.h is not included at the top of the file. This PR adds it.


sletz:
Thanks.

PR #341, opened 2019-08-04
Double the Machine Pointer Size when under 32bit Android

Jman420:
Make llvm-test temp directory configurable


sletz:
Thanks, but not sure the gMachinePtrSize fix will work on Windows and so on . So can you make it Android specific for now? Thanks.

PR #352, opened 2019-10-06
Fix potential memory leak when creating OpenSL ES engine

selimbousbih:
Memory leak occurs during init when I switch between different dsp objects multiple times on the same DspFaust instance.

### DIFF
--- a/architecture/faust/audio/android-dsp.h
+++ b/architecture/faust/audio/android-dsp.h
@@ -341,6 +341,26 @@ class androidaudio : public audio {
                     return false;
             }
           
+			if (fInputBufferQueue) {
+				(*fInputBufferQueue)->Destroy(fInputBufferQueue);
+				fInputBufferQueue = NULL;
+			}
+
+			if (fOutputBufferQueue) {
+				(*fOutputBufferQueue)->Destroy(fOutputBufferQueue);
+				fOutputBufferQueue = NULL;
+			}
+
+			if (fOutputMix) {
+				(*fOutputMix)->Destroy(fOutputMix);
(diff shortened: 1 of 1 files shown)


sletz:
Coud you send an example code that shows the problem in the first place? Thanks.

selimbousbih:
This is how I got to the issue (I'm testing on android)
```
// C++
// override of FaustPolyEngine init method
void init(int dsp_id, audio *driver, midi_handler *midi){
    MyDspFactory *myDspFactory = new MyDspFactory();
    dsp *mono_dsp = myDspFactory->create(dsp_id);
	
    // ...
}

void refresh(int id){
    init(id, fDriver, &fMidiHandler);
}

// Java
void selectDsp(int id){
    dspFaust.stop();
    dspFaust.refresh(id);
}

```

After the 14th call of refresh here is what I get
D/JNI: DspFaust::refresh
E/Faust: init
E/libOpenSLES: Too many objects
W/libOpenSLES: Leaving Engine::CreateAudioPlayer (SL_RESULT_MEMORY_FAILURE)
E/Faust: CreateAudioPlayer/openSLEngineInterface error  = Memory failure
    Destructor

A new instance of fOpenSLEngine is always created after each init call but it is only freed when the driver itself is destructed (In this case androidaudio driver)

sletz:
OK, I reworked you PR a bit and committed the result in https://github.com/grame-cncm/faust/commit/7e38a195c58c2ab2268fa1aa4a573cf4e76f9e5d.
Can you test and report? Thanks.

selimbousbih:
Works fine, thanks.

PR #367, opened 2019-12-02
Better guess at Pd's include path on macos, and allow override with env variable

giuliomoro:
This makes `faust2puredata` usable again


agraef:
Hmm yes, Pd-extended has definitely bitten the dust. ;-) I've solved this upstream by just including a recent version of m_pd.h in the sources, that's all that puredata.cpp needs.

PR #372, opened 2019-12-12
support AudioParams change with Firefox compatibility

Fr0stbyteR:


sletz:
Thanks. 
I think a  destroy method (as in https://github.com/grame-cncm/faust2webaudio/pull/5/commits/21b153e3121a508b340677fa54a9f29c20ad4588) should also be added.

PR #374, opened 2019-12-21
Fix "make benchmark"

kmatheussen:


sletz:
Thanks, did that manually.

PR #386, opened 2020-01-15
Initial commit to make the build process (install, in fact) compatible w/ Gentoo FHS

rugubara:
Initial attempt to make this project comply with gentoo FHS.

Gentoo requires that 32-bit libs are located in /usr/lib but 64-bit libs are located in /usr/lib64.

The project hard-codes the lib location to ${PREFIX}/lib

This commit introduces INSTALL_LIBDIR variable that allows to fix all lib locations
(with default set to ${PREFIX}/lib) except
usr/lib/libfaustmachine.so.1.0.0
usr/lib/libOSCFaust.so.1.2.2

I seek advice from upstream to get these 2 fixed.


dfober:
Does it work for you ? INSTALL_LIBDIR is declared in the makefile and used by cmake. The normal way is to pass it to cmake using a -D option e.g. cmake -DINSTALL_LIBDIR=lib64.

rugubara:
@dfober , I didn't quite catch your question. 
Gentoo ebuild dev guide requires me to evaluate the lib install dir in the ebuild compile and install phases using the provided function get_libdir (see [Project:AMD64](https://wiki.gentoo.org/wiki/Project:AMD64/Multilib_layout#Programs_installing_64-bit_libraries_to_.2Fusr.2Flib_.28not_respecting_libdir.29) ) 
I export  INSTALL_LIBDIR=$(get_libdir) and call **make world**.  I want to propagate this setting to the build process.

As I did my modifications, all targets obeyed except usr/lib/libfaustmachine.so.1.0.0 and usr/lib/libOSCFaust.so.1.2. Can you help me to identify the places in Makefiles and CMakeFiles where they are built so that I could include them into this PR?

dfober:
ok, I'll have a look.

dfober:
I'll start with your changes and I'll let you know when it's done.

dfober:
That's done. You can update your repo. It was not straightforward: faust exe needs to know the libraries install location (due to the --libdir option). Read the readme and/or type make help to use. This is also documented in the [faust wiki](https://github.com/grame-cncm/faust/wiki/install).

PR #391, opened 2020-01-31
Esp32

thopman:
Fixed esp32 midi implementation. 

added component.mk and CMakeLists.txt to use the codec drivers and generated FaustDSP as esp-idf components. 

Added extra options for midi and for the future polyphony, in the faust2esp32 script. When using the -lib option it generates files like before. if you leave out -lib option it will compile and generate the files as esp-idf components. faust2esp32 -h updated to reflect this. 

jdksMidi is not included, because i will put this in a seperate repository. Will probably write some midi parse code myself to replace this.


sletz:
Did a bit of (blind...) cleanup and merged. Can you test again? Thanks !

PR #396, opened 2020-02-03
esp32_midi():midi_handler("esp32"):fProcessMidiHandle(NULL) throws a …

thopman:
…compile error. Added NULL initialization to where the taskhandle gets declared.

+

Removed duplicate fMIDIHandler -> startMidi() and stopMidi() calls.


thopman:
They get called in the midiui class. fMIDIInterface->run() and stop();

sletz:
Manually done, thanks.

PR #401, opened 2020-02-13
Implemented "NVOICES" poly mode

thopman:
Fixed a bug in esp32-midi.h and raised the available task stack size.


sletz:
Cleanup  and merged, thanks.

PR #405, opened 2020-02-27
Use `MediaDevices` instead of deprecated API for `getUserMedia`

Korilakkuma:
Use `MediaDevices` because of `navigator.getUserMedia` is deprecated API ([NOTE](https://www.w3.org/TR/mediacapture-streams/#legacy-interface-extensions)).


sletz:
Thanks!

PR #412, opened 2020-03-05
Clean up documentation on compilation

### LINKED ISSUE #408: Documentation on compiling libfaust
Hi! I wanted to compile libfaust to use it in my application, but it was hard
to get started because I found the documentation a bit confusing.

Looks like the instructions for compiling Faust (and libfaust in particular)
are spread out across at least three places:

- The `compiling` section in the manual
- The `build/README.md` file
- The wiki

These three places seem to have more or less the same information, but with
some minor differences, and also some outdated information (they mention commands
like `make configstatic`, which don't exist in the makefiles anymore). It was
(and to some extent still is) unclear to me which of these guides to follow.
The documentation should be all in one place.

The wiki seems to have the most comprehensive information. What is its role in
relation to the manual? If the wiki is the right place for the build
instructions, I think the manual and `build/README.md` should be updated to
just point to it. The manual could have maybe a few sentences on the basic
compilation, but I think having the detailed instructions there (related to
MSVC and whatnot) is unnecessary.

If this sounds right, I could make the above changes to the documentation and
README, and try to write them so that they are easy to follow for someone who is not very
familiar with the Faust ecosystem (like me).

mgunyho:
This is resolves #408.

I removed some duplicate information across the documentation related to compilation. [Here](https://github.com/grame-cncm/faust/files/4295088/wiki-pages.zip) are markdown files corresponding  to wiki pages with some changes.

- I removed all detailed compilation instructions from the main `README` and `build/README.md` and made them link to the wiki instead.
- I changed links saying "for detailed instructions, see `build/README`" to point to the wiki
- The first part of the compilation section in the manual is left mostly intact (so there is a bit of duplicate info), but I removed the more advanced instructions and added links to the wiki.
- The MSVC/MSYS2 specific instructions are still in `build`, I added links to those in the wiki pages
- There was a section 'Notes regarding the backends compilation' in the `build/README` that wasn't in the wiki, I added it to the `backends` wiki page. Not sure how relevant or up to date it is.
- I also added some small changes to the `embedding` manual page

I tried to make sure that everything that was removed from the MD's is in the wiki files.

I'm not that familiar with CMake so if you find any mistakes, please feel free to correct them.


dfober:
That's fine! Thanks!

Issue #445, opened 2020-06-07
Rust backend produces unexpected output for physical model generators


bluenote10:
When experimenting with this simple Faust dsp
```faust
import("stdfaust.lib");

gate = 1.0;
freq = 440.0;

process = pm.elecGuitar(
  pm.f2l(freq),
  hslider("pluck position", 0.9, 0, 1, 0.01),
  hslider("mute", 1.0, 0, 1, 0.01),
  hslider("pluck gain", 0.8, 0, 1, 0.01),
  gate
);
```
I'm not quite getting the expected output from the Rust backend.

Testing this Faust code in the online IDE suggest that it is correct in general, producing a single note output. With the Rust backend I'm only getting 108 non-zero output samples, then the signal dies to zero, which is essentially just a short clicking noise.

I have prepared a minimal architecture file I'm using to debug the Rust output:

<details>
<summary>Debug architecture file</summary>

```rust
#![allow(unused_parens)]
#![allow(non_snake_case)]
#![allow(non_camel_case_types)]
#![allow(dead_code)]
#![allow(unused_variables)]
#![allow(unused_mut)]
#![allow(non_upper_case_globals)]

extern crate libm;

use std::fs::File;
use std::io::Write;

pub trait Meta {
    // -- metadata declarations
    fn declare(&mut self, key: &str, value: &str) -> ();
}

pub trait UI<T> {
    // -- widget's layouts
    fn openTabBox(&mut self, label: &str) -> ();
    fn openHorizontalBox(&mut self, label: &str) -> ();
    fn openVerticalBox(&mut self, label: &str) -> ();
    fn closeBox(&mut self) -> ();

    // -- active widgets
    fn addButton(&mut self, label: &str, zone: &mut T) -> ();
    fn addCheckButton(&mut self, label: &str, zone: &mut T) -> ();
    fn addVerticalSlider(&mut self, label: &str, zone: &mut T, init: T, min: T, max: T, step: T) -> ();
    fn addHorizontalSlider(&mut self, label: &str, zone: &mut T , init: T, min: T, max: T, step: T) -> ();
    fn addNumEntry(&mut self, label: &str, zone: &mut T, init: T, min: T, max: T, step: T) -> ();

    // -- passive widgets
    fn addHorizontalBargraph(&mut self, label: &str, zone: &mut T, min: T, max: T) -> ();
    fn addVerticalBargraph(&mut self, label: &str, zone: &mut T, min: T, max: T) -> ();

[...]

PR #446, opened 2020-06-08
Basic impulse tests for rust

bluenote10:
As suggested in #445, here is a basic setup for running the impulse tests with Rust.

Currently the tests are still failing, which is most likely because I'm not fully replicating the logic of the impulse tests yet. In particular, I haven't fully understood the differences of the 4 modes, and for now I'm just running the impulse mode 4 times. At least many tests match in the first 30000 samples already, so this should work as a foundation to play around with the tests.

A few implementation notes:

- I decided to use a `cargo run --bin <DSP>` to compile and run the Rust code. Therefore the tests are executed within that empty cargo project skeleton.

- The binaries don't write to stdout directly, but rather take an output file path as command line argument which is used for writing the IR data.

- At first I cached the generated Rust sources on the makefile level similar to how it is done for C++, but this turned out to slow down development, because with every change to the architecture file, you'd have to clean the targets to force a re-run. Since the Faust code generation is so fast I figured, there is no need for the caching. To keep things simple, the generation + compile + run + compare is now a single step target in the makefile. The execution looks like that:

    image

- I removed the unused existing Rust architecture file to avoid confusion.

- I'm also went for running with `-double`, but I noticed that some of the generated code doesn't compile (e.g. `cubic_distortion.dsp`) because of missing `f32` / `f64` casts.


bluenote10:
I think I've found something: The most suspicious test case related to #445 is the `comb_delay1.dsp`, because it also shows a deviation after a similar number of samples:

```
./filesCompare ir/rust/comb_delay1.ir reference/comb_delay1.ir 
line : 109 output : 0 sample1 : 0 different from sample2 : 0.062565 delta : 0.062565
line : 110 output : 0 sample1 : 0 different from sample2 : 0.124979 delta : 0.124979
line : 111 output : 0 sample1 : 0 different from sample2 : 0.186903 delta : 0.186903
line : 112 output : 0 sample1 : 0 different from sample2 : 0.248092 delta : 0.248092
line : 113 output : 0 sample1 : 0 different from sample2 : 0.308306 delta : 0.308306
line : 114 output : 0 sample1 : 0 different from sample2 : 0.367308 delta : 0.367308
line : 115 output : 0 sample1 : 0 different from sample2 : 0.480671 delta : 0.480671
line : 116 output : 0 sample1 : 0 different from sample2 : 0.534674 delta : 0.534674
line : 117 output : 0 sample1 : 0 different from sample2 : 0.63617 delta : 0.63617
line : 118 output : 0 sample1 : 0 different from sample2 : 0.683266 delta : 0.683266
line : 119 output : 0 sample1 : 0 different from sample2 : 0.769226 delta : 0.769226
line : 120 output : 0 sample1 : 0 different from sample2 : 0.843054 delta : 0.843054
```

Looking at the generated code I've found that the "shifting" of the vector content is the problem:

```rust
for j0 in 11..0 {
    self.fVec0[j0 as usize] = self.fVec0[(j0 - 1) as usize];
}
```

That is a bit of a gotcha of Rust ranges: The `11..0` range doesn't really count down, but is an empty (upwards counting) range. To make it count down, one can use `rev` on the range. By replacing the code with the following I can make the IR test identical (in the first 30000 samples that is):

```rust
for j0 in (1..=10).rev() {
    self.fVec0[j0 as usize] = self.fVec0[(j0 - 1) as usize];
}
```

sletz:
OK, thanks. I'll have a look at the code generator on this specific case then.

sletz:
And the [0...N] range has to be changes to [1..N-1] right ?

sletz:
Can you check the fix in https://github.com/grame-cncm/faust/commit/0b6d56ad866ca9676bfcd7c4a36c187c22f94461 and report?

bluenote10:
> And the [0...N] range has to be changed to [1..N-1] right?

Yes, either `(1..=10).rev()` (inclusive upper value) or `(1..11).rev()` (exclusive upper value) should work.

bluenote10:
> Can you check the fix in 0b6d56a and report?

Thanks for the quick fix! That looks good, after merging the fix into the test branch the `comb_delay1` test now passes the first 30000 samples:

```
./filesCompare ir/rust/comb_delay1.ir reference/comb_delay1.ir 
line : 30109 output : 0 sample1 : 0.062565 different from sample2 : 0.250259 delta : 0.187694
line : 30110 output : 0 sample1 : 0.124979 different from sample2 : 0.499918 delta : 0.374939
line : 30111 output : 0 sample1 : 0.186903 different from sample2 : 0.747612 delta : 0.560709
line : 30112 output : 0 sample1 : 0.248092 different from sample2 : 0.992367 delta : 0.744275
line : 30113 output : 0 sample1 : 0.308306 different from sample2 : 1.23322 delta : 0.924917
line : 30114 output : 0 sample1 : 0.367308 different from sample2 : 1.46923 delta : 1.10192
line : 30115 output : 0 sample1 : 0.480671 different from sample2 : 1.92268 delta : 1.44201
line : 30116 output : 0 sample1 : 0.534674 different from sample2 : 2.13869 delta : 1.60402
line : 30117 output : 0 sample1 : 0.63617 different from sample2 : 2.54468 delta : 1.90851
line : 30118 output : 0 sample1 : 0.683266 different from sample2 : 2.73306 delta : 2.0498
line : 30119 output : 0 sample1 : 0.769226 different from sample2 : 3.0769 delta : 2.30768
line : 30120 output : 0 sample1 : 0.843054 different from sample2 : 3.37221 delta : 2.52916
```

Looking at 0b6d56a, I'm wondering if it might be an issue that `genSimpleForLoopInst` is also used by `generateInitArray` and `generateCopyArray`, which may now miss to iterate over the zeroth element?

How should we proceed with the test suite? The particular questions I have:
- The sample range 15000 to 30000 seems to feed random values into the DSPs. Is this deterministic in the C++ tests by fixing the seed? In this case, the challenge on Rust side would be to re-implement `rand()` exactly to match the sequence. Maybe the random sequence could be provided externally, or we use a simple pseudo-random sequence that is easier to replicate?
- The range starting at 30000 seems to deal with wrapping the DSP into polyphonic mode, which does not exist for Rust. Is there is simple way to emulate the behavior for the impulse tests? Or should we alternatively give `filesCompare` a command line argument up to which sample number it should compare, so that the Rust IR tests could omit the checks on polyphonic modes.

sletz:
- **which may now miss to iterate over the zeroth element?**  The code generation for revers is different here: https://github.com/grame-cncm/faust/blob/master-dev/compiler/generator/rust/rust_instructions.hh#L449
- **The sample range 15000 to 30000 seems to feed random values into the DSPs.**  yes feeding a fixed list of vela would be better.
- **DSP into polyphonic mode** : I plan to work on your PR now. Can you integrate the latest fix of master-dev it it?

bluenote10:
> I plan to work on your PR now. Can you integrate the latest fix of master-dev it it?

Awesome, already pushed.

sletz:
- PR improved an bit and merged in https://github.com/grame-cncm/faust/commit/4b6bed36e1fbfb02a14925c3ef786b2d9d02d416, thanks !
- only one test really fails (reverb_designer.dsp) the other failing ones fail because the testing procedure itself is not completely correct
- I could not find the math `rint` function. Using the `linux_api_math::rint ` fills because the linux_api_math create cannot be found. Any idea here?
- now hoping your initial https://github.com/grame-cncm/faust/issues/445 will work. Please report !

bluenote10:
For clarity: PR has been merged manually => closing.

> I could not find the math rint function. Using the linux_api_math::rint fails because the linux_api_math create cannot be found. Any idea here?

Any reason why a native Rust solution like `x.round() as i32` would not be sufficient? Anyhow we can deal with that in a next step.

Thanks again for all the fixes.

sletz:
We have `rint` and `round`  in the language.

bluenote10:
What is the difference between the two? If it is only about bankers rounding vs arithmetic rounding, this can probably by expressed directly in Rust.

sletz:
Rust backend fixed and test completed in https://github.com/grame-cncm/faust/commit/86630c5a34461a6045975bd370eec6fbee113970. The one failing one is related with this `rint` issue.  Now Rust backend use can carry on. Please report some interesting use-cases !

Issue #409, opened 2020-03-04
Rust: `UI<T>::declare()` and borrow


bkfox:
Rust code provides `UI<T>::declare(&mut self, zone: &mut T, key: &str, value: &str) -> ()` used to declare variables accessible from user interface.

The parameter `zone` is a mutable borrow to member variable of the generated DSP, which is not really usable: an UI implementation can't keep the provided member reference in order to update it when a control event occurs, because of borrow and ownership rules.

Potential solution would assign an index for each generated variable that can be used to retrieve or update members' values. On the generated DSP struct, two accessor methods: `fn get_var(&self, index: Index) -> &f32`, `fn get_var_mut(&mut self, index: Index) -> &mut f32` returning actual reference to DSP's variables by index. This index is passed down to `UI::declare()` as the `zone` instead of the reference.
Although there still is an undefined part on how concrete implementation of UI access the DSP instance (and more or less related thing is the ownership model used here), it leaves users freedom to choose how to integrate DSP and UI in their applications.

sletz:
TBH the Rust backend is still very experimental. So I understand you propose to adapt the https://github.com/grame-cncm/faust/blob/master-dev/architecture/minimal.rs file and the generated code? Can you prepare a pull-request ?

sletz:
There is also something to keep in mind: in the C++ model several UI can be connected to a same DSP so that each of them can possibly update the `zones`. Does your proposal solve this issue?

bkfox:
Sorry I could not answer sooner. I'm actually working on a similar problem for an audio project using a DSP graph, where multiple UI should be able to update dsp's parameters. I'll however make more progress into this direction before I actually can give you a more concrete answer.

Here are some thoughs:
- DSP reference passed to `UI::declare` should be generic and passed down to `buildUserInterface` from the caller. This avoid to force generated DSP to take care of such implementation detail as how to refer to themselves and opens possibilities to multiple UI backends.
- Using variables references for zones poses the problem of ownership, lifetime and multithread (maybe): multiple UI keeping direct reference to a DSP variable means that they can't outlive the DSP. A UI would probably keep theses references in a map, such as:

    ```rust
    UI<'a> {
        zones: BTreeMap<&'a mut f32, Metadata>,
        //... 
    } 
    ```

    Where it becomes difficult it when UI has to handle multiple DSP.
[...]

PR #448, opened 2020-06-19
[WIP] add idiomatic dsp trait for rust

bluenote10:
As a first step for #409 this PR tries to add a trait for underlying the Rust code generation. So far, no changes regarding UI interfaces are included. I guess it's better to leave that as a second step. The fundamental trait would look like:

```rust
pub trait FaustDsp {
    type Float;

    fn new() -> Self where Self: Sized;
    fn metadata(&mut self, m: &mut dyn Meta);
    fn get_sample_rate(&mut self) -> i32;
    fn get_num_inputs(&mut self) -> i32;
    fn get_num_outputs(&mut self) -> i32;
    fn get_input_rate(&mut self, channel: i32) -> i32;
    fn get_output_rate(&mut self, channel: i32) -> i32;
    fn class_init(sample_rate: i32) where Self: Sized;
    fn instance_reset_user_interface(&mut self);
    fn instance_clear(&mut self);
    fn instance_constants(&mut self, sample_rate: i32);
    fn instance_init(&mut self, sample_rate: i32);
    fn init(&mut self, sample_rate: i32);
    fn build_user_interface(&mut self, ui_interface: &mut dyn UI<Self::Float>);
    fn compute(&mut self, count: i32, inputs: &[&[Self::Float]], outputs: &mut[&mut[Self::Float]]);
}
```

**Notes**

- The trait is object safe. I've adapted the impulse tests so that they actually verify object safety. For instance, if you remove one of the `where Self: Sized` annotations (a object safety violation), the code would no longer compile.
- In order to make it easier to support both `f32` and `f64` dsp types, I have introduced that underlying float type as an associated type the to trait. Otherwise we would either have to introduce two separate traits for `FaustDsp32` and `FaustDsp64` or the trait definition from an architecture file would only be compatible with either `f32` or `f64` code generation. With the associated type it is possible for a project to have mixed DSPs in f32 and f64 as well. In use cases where only a single dsp type is needed (more common I guess), users can easily simplify things with a type alias like `type Dsp = FaustDsp<Float=f32>`. I went for the name `FaustDsp` to encourage such type aliases and make it easier for project to use the name `Dsp` as desired.
- In order to make the public interfaces more idiomatic, I tried to adapt the code generation to produce snake-case function names.
- I'm generating all methods under the `impl FaustDsp for MyDsp { ... }` scope now, i.e., there is not directly `impl MyDsp { ... }` at all, which simplifies things. If it turns out that we need both, the code generation would probably require a refactoring, because we would have to sort functions depending on where they should get attached.
- I'm only using the trait for the "class" but not the "internal class" generation. I don't understand the role of the internal class yet, so I'm not quite sure if using the trait would be preferred there as well.
- I don't fully understand yet why the code generates the unnecessary `fDummy` field. I feel like it might be related to the associated type. Maybe now that the type is encoded in the trait already we can get rid of it?

**Status** 

Currently many impulse tests already work, but there are some problems. Actually, all failures seem to have the same cause at the moment (e.g. `comb_delay1`). In these cases, two things happen:

- My attempt at `produceInfoFunctions` actually doesn't produce any code for some reason, leading to the error `not all trait items implemented, missing: get_num_inputs, get_num_output`, get_input_rate, get_output_rate`. Why does the code generation sometimes omit them? To satisfy the trait requirements it would be necessary to produce them unconditionally.
- The `generateStaticInit` produces a line over which I don't have control, which produces an invalid function call name e.g. `sig0.instanceInitmydspSIG0(sample_rate);`. This seems to happen exactly if and only if the `produceInfoFunctions` omits the functions.

CC @bkfox Would be good to have your opinion on the trait design.

### DIFF
--- a/architecture/minimal-jack.rs
+++ b/architecture/minimal-jack.rs
@@ -26,6 +26,8 @@ use jack::prelude as j;
 use std::io;
 extern crate libm;
 
+// TODO: adapt
+
 pub trait Meta {
 
     // -- metadata declarations
--- a/architecture/minimal-portaudio.rs
+++ b/architecture/minimal-portaudio.rs
@@ -26,6 +26,8 @@ use portaudio as pa;
 use std::io;
 extern crate libm;
 
+// TODO: adapt
+
 pub trait Meta {
 
     // -- metadata declarations
--- a/compiler/generator/rust/rust_code_container.cpp
+++ b/compiler/generator/rust/rust_code_container.cpp
@@ -103,7 +103,7 @@ void RustCodeContainer::produceInternal()
     *fOut << "pub struct " << fKlassName << " {";
 
     tab(n + 1, *fOut);
-   
+
     // Fields
     fCodeProducer.Tab(n + 1);
     generateDeclarations(&fCodeProducer);
@@ -117,14 +117,14 @@ void RustCodeContainer::produceInternal()
 
     tab(n + 1, *fOut);
     tab(n + 1, *fOut);
-    produceInfoFunctions(n + 1, fKlassName, "&mut self", false, false, &fCodeProducer);
+    produceInfoFunctions(n + 1);
 
     // Init
     // TODO
     // generateInstanceInitFun("instanceInit" + fKlassName, false, false)->accept(&fCodeProducer);
 
     tab(n + 1, *fOut);
-    *fOut << "pub fn instanceInit" << fKlassName << "(&mut self, sample_rate: i32) {";
+    *fOut << "pub fn instance_init" << fKlassName << "(&mut self, sample_rate: i32) {";
     tab(n + 2, *fOut);
     fCodeProducer.Tab(n + 2);
     generateInit(&fCodeProducer);
@@ -190,7 +190,7 @@ void RustCodeContainer::produceClass()
 
     *fOut << "pub struct " << fKlassName << " {";
     tab(n + 1, *fOut);
-   
+
     // Dummy field used for 'declare'
     *fOut << "fDummy: " << ifloat() << ",";
     tab(n + 1, *fOut);
@@ -204,7 +204,7 @@ void RustCodeContainer::produceClass()
     tab(n, *fOut);
 
     tab(n, *fOut);
-    *fOut << "impl " << fKlassName << " {";
+    *fOut << "impl FaustDsp for " << fKlassName << " {";
 
     // Memory methods
     tab(n + 2, *fOut);
@@ -229,7 +229,7 @@ void RustCodeContainer::produceClass()
         tab(n + 1, *fOut);
     }
 
-    *fOut << "pub fn new() -> " << fKlassName << " { ";
+    *fOut << "fn new() -> " << fKlassName << " { ";
     if (fAllocateInstructions->fCode.size() > 0) {
         tab(n + 2, *fOut);
         *fOut << "allocate" << fKlassName << "(dsp);";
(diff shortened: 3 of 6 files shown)


sletz:
The falling cases all use table (rtable, rwtable) that generate sub classes. I'll have a look.

bkfox:
Well done for this work :)

Here are just a few notes:
- Float associated type: why not even use a generic numeric value (as `Sample`)? Although it does not seems to be the vibe of Faust to handle integer, this would be good to have it in the trait IMHO. 
- Some methods take a mutable `self` on different methods -- which may not be required: `get_sample_rate`, `metadata`, `get_num_outputs`, `get_num_inputs`, `build_user_interface`;

sletz:
> * Float associated type: why not even use a generic numeric value (as `Sample`)? Although it does not seems to be the vibe of Faust to handle integer, this would be good to have it in the trait IMHO (this would open a way for a bright future of handling generic Sample type in DSP implementation)

Faust currently handles integer (32 bits) or float/double types for signals. We could also use REAL instead of Float and Sample. This is was is currently done in some C++ architectures that use C++ template to switch between float/double at compile time, see for instance: https://github.com/grame-cncm/faust/blob/master-dev/architecture/faust/gui/UI.h

> * Some methods take a mutable `self` on different methods -- which may not be required: `get_sample_rate`, `metadata`, `get_num_outputs`, `get_num_inputs`, `build_user_interface`;

Yes, good point. Those methods do not change the DSP internal state, and typing them correctly would be better yes.

> * I'm really not sure about it, but maybe `get_num_input` and `get_num_output` can return an `Option` **if** Faust allows to handle any number of channels as input (which will result in the same output channels' count);

Nope. Faust compiler generates completely "static" code, so inputs/outputs are known at compile time.

sletz:
_Correct Rust code generator, all impulse-tests working again, make faust2jackrust and faust2portaudiorust work again._
Then merged in rebase mode. 
Thanks!

bluenote10:
Awesome, thank you very much! I will soon start working on the next steps for UI compatibility.

> Some methods take a mutable self on different methods

Fully agreed, I still had that on my todo list.

> Float associated type: why not even use a generic numeric value (as Sample)?

> We could also use REAL instead of Float and Sample.

Just renaming it to `Sample` would be fine by me. The name `Real` is a bit of a misnomer, that we should maybe avoid. I've seen it mainly used in C/C++ contexts as a consequence of the historically bad float vs double naming convention. A "real" number is mathematically exactly what a floating point number cannot represent. Most modern programming languages including Rust use the term "float" to refer to floating point numbers of any precision, which makes sense, because that is their main property, and using `Real` would not really be idiomatic in Rust.

@bkfox If you don't mean renaming, but switching from an associated type to actual trait-based generics (like `num::traits::Float` or even something more general), we have to be careful. So far my experience with using `num::traits::Float` is not so great, because its interface is sometimes too broad, i.e., you cannot do everything with it you can do with `f32` or `f64` directly. Floating point math is messy enough for a specific type already, and trying to abstract over all the quirks generically doesn't make it easier. Since the Faust compiler makes fairly specific assumptions about how `f32` or `f64` should behave, I'd be worried if a trait aligns well with these assumptions. Other issues with `num::traits::Float` is that it complicates debugging, and I've had cases where the Rust compiler failed to optimize away type conversions. I'd say an associated type gives the flexibility needed to work with at least `f32` and `f64` seamlessly while avoiding the complexity of really having to support arbitrary numeric types.

PR #454, opened 2020-06-25
[Rust] Index based parameter interface

bluenote10:
The second part to solve #409. Basically the idea is that the generator now produces:

```rust
fn build_user_interface(&self, ui_interface: &mut dyn UI<Self::Sample>) {
    Self::build_user_interface_static(ui_interface);
}

fn build_user_interface_static(ui_interface: &mut dyn UI<Self::Sample>) {
    ui_interface.open_vertical_box("APF");
    ui_interface.add_horizontal_slider("Freq", ParamIndex(0), 1000.0, 100.0, 10000.0, 1.0);
    ui_interface.add_horizontal_slider("Q", ParamIndex(1), 1.0, 0.01, 100.0, 0.01);
    ui_interface.close_box();
}

fn get_param(&self, param: ParamIndex) -> Option<Self::Sample> {
    match param.0 {
        0 => Some(self.fHslider0),
        1 => Some(self.fHslider1),
        _ => None,
    }
}

fn set_param(&mut self, param: ParamIndex, value: Self::Sample) {
    match param.0 {
        0 => { self.fHslider0 = value }
        1 => { self.fHslider1 = value }
        _ => {}
    }
}
```

This should allow to build UIs without having to worry about shared mutability: Via the `build_user_interface` functions, the UI can learn which parameters have which index, and from there it can can use the getter/setter on the DSP to do its thing.

@bkfox Would that do the job for you?

Notes:

- As discussed in the issue, I've added both a static and a non-static user interface builder (all object safe).
- When I was reading #409 for the first time I couldn't follow so well because the term "zone" confused me. The problem is that it has the connotation of "range" or "area", and that is not really what the UI controls: For instance, having `set_zone` or `get_zone` sounds like they always refer to two values, a "from" and a "to" value, which isn't the case. It my opinion it would be clearer to speak of "dsp parameters" in general (which may have a "zone" associated with them in the case of slider, but it is not really what is read/written). Would everyone be happy with the proposed naming scheme?
- Similarly the name `instance_reset_user_interface` wasn't quite fitting, because this function technically resets the parameters, and it makes even sense to call it without having any user interface. My suggestion would be to call it `instance_reset_params` to make it more generally applicable.
- The change allowed to get rid of `fDummy` as well: `declare` now simply takes a `param: Option<ParamIndex>` argument, which allows to handle the nullptr case.
- It is slightly weird that the parameter getter/setter operate with `Self::Sample`. Would you prefer calling it just `Self::T` for brevity/generality?
- Implementation wise what was needed is an extra iteration over the "user interface instruction" to build up the parameter to index lookup table. At first I did the building of the lookup table and the code generation in one go, but I thought it is better to go for a single responsibility approach and have two visitors, to make it more obvious that also the getter/setter generation requires the lookup table.


sletz:
> @bkfox Would that do the job for you?
>

That would be cool to have a working demo that shows how several UI can access/modify a sample DSP instance.

> * Similarly the name `instance_reset_user_interface` wasn't quite fitting, because this function technically resets the parameters, and it makes even sense to call it without having any user interface. My suggestion would be to call it `instance_reset_params` to make it more generally applicable.

Right, but I'm a bit annoyed by the naming change we introduce compared to other backends. Or should we fix all of them with this better name ? ((-; 

> * The change allowed to get rid of `fDummy` as well: `declare` now simply takes a `param: Option<ParamIndex>` argument, which allows to handle the nullptr case.
> * It is slightly weird that the parameter getter/setter operate with `Self::Sample`. Would you prefer calling it just `Self::T` for brevity/generality?

So you mean renaming `Sample` to `T` evert where is appears ?

bluenote10:
> Right, but I'm a bit annoyed by the naming change we introduce compared to other backends.

I also wouldn't mind reverting it, I just looked at it from the context of producing a nice interface in Rust. One may argue that backends may require language specific deviations anyway to result in idiomatic interfaces and to incorporate special language features like Rust's borrow checker. In that sense the Rust backend is probably the only one that introduces the notion of "set_param" / "get_param", which makes "reset_params" more fitting than for other backends.

> So you mean renaming Sample to T evert where is appears ?

Yes, just an idea. The good old generic `T` just fits for both sample and parameters, and on the `UI` trait we call it `T` as well so far (but since the `UI` trait can be provided by the user its generic parameter doesn't matter as much).

sletz:
> I also wouldn't mind reverting it, I just looked at it from the context of producing a nice interface in Rust. One may argue that backends may require language specific deviations anyway to result in idiomatic interfaces and to incorporate special language features like Rust's borrow checker. In that sense the Rust backend is probably the only one that introduces the notion of "set_param" / "get_param", which makes "reset_params" more fitting than for other backends.
>

OK, so lets kept that for now.

> > So you mean renaming Sample to T evert where is appears ?
> 
> Yes, just an idea. The good old generic `T` just fits for both sample and parameters, and on the `UI` trait we call it `T` as well so far (but since the `UI` trait can be provided by the user its generic parameter doesn't matter as much).

T seems better, so please correct , thanks !

bluenote10:
> T seems better, so please correct , thanks !

Okay, done!

sletz:
Thanks. 

Since this development is somewhat experimental, I will push this branch on a new "reworked_rust_backend" on the GRAME-CNCM GitHub, so that we can work on it until is is stable enough to be merger back in master-dev. OK for you?

bluenote10:
Sure, it anyway makes sense to wait and hear @bkfox opinion if that resolves all aspects of #409.

bluenote10:
For what it's worth, I did some pretty extensive testing on this, and so far I'm pretty happy from the Rust user side. All my experiments with buildings UIs worked fine so far. I hope I can soon share a small demo or so.

Apparently @bkfox is a bit busy, but considering that the interface that is on `master-dev` does not work at all for building UIs, we might as well bring this back?

jmyhrman:
I really like the ergonomics from this approach of getting/setting values. Could we have this rebased on lastest master-dev to evaluate further? Or what would be necessary to bring it to master-dev?

bluenote10:
If I remember correctly, we had created a special branch (I think it was `rust-wip`) where this has been merged into.

In my opinion it would make sense to bring `rust-wip` back to `master-dev` though because:
- It isn't so much WIP anymore. It works pretty stable as it is for me.
- What is on `master-dev` doesn't work at all for building UIs, so it is a step forward anyway.
- The extra branch is a bit inconvenient / confusing for newcomers.

sletz:
Yes, I'll have a look ASAP

sletz:
Rebase + merge done in https://github.com/grame-cncm/faust/commit/c025415420ac6542292d0b837c466ecb15b5ea4d
Impulse-tests and `faust2jackrust`/`faust2portaudiorust` working.

bluenote10:
> Rebase + merge done in c025415

Cool thanks! Does it mean we can close the PRs now because they are all merged manually?

sletz:
Yes, thanks.

PR #459, opened 2020-06-29
Enable AudioUnit's input only when fDevNumInChans > 0

### LINKED ISSUE #457: faust2api -ios: How to build an app not asking for microphone access?
The iOS application I am creating using `faust2api` is a synthesizer not currently using audio inputs at all.

By default, when launched for the first time, the application asks the user to provide it with Microphone access.

It still works fine no matter you allow or disallow, however it affects the overall first launch experience anyway.

I'm wondering if there is a way to build the C++ API in a way that wouldn't require microphone access?

Thanks in advance.

oshibka404:
closes #457

The issue seems to be somehow related to this: https://stackoverflow.com/questions/21337276/ios-app-asks-for-microphone-access-even-for-kaudiosessioncategory-mediaplayback
and this: alexbw/novocaine#76

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Faust - Programming Language for Audio Applications and Plugins
 
-## Grame, Centre National de Creation Musicale: <http://www.grame.fr>
+## Grame, Centre National de Creation Musicale: <https://www.grame.fr>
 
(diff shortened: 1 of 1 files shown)


sletz:
Thanks a lot!

PR #463, opened 2020-07-10
updated inkscape command line option to work with Inkscape 1.0

LFSaw:
Inkscape version > (4035a4f, 2020-05-01) does not include `-o` option anymore, but uses `-E` instead.

### DIFF
--- a/tools/faust2appls/faust2pdf
+++ b/tools/faust2appls/faust2pdf
@@ -8,7 +8,7 @@ for d in $@; do
         name=${d%.dsp}
         cd $name-svg
         for f in *.svg; do
-            inkscape -E ${f%.svg}.eps $f
+            inkscape -o ${f%.svg}.eps $f
             epstopdf ${f%.svg}.eps
             rm $f
         done


LFSaw:
I downloaded inkscape from their [release](https://inkscape.org/release/inkscape-1.0/) page and used the binary found in `Inkscape.app/Contents/MacOS`.

from its man-page:

```
File export:
  -o, --export-filename=EXPORT-FILENAME      Output file name (file type is guessed from extension)
```


This version 
```bash
$ ./inkscape --version
Inkscape 1.0 (4035a4f, 2020-05-01)
    Pango version: 1.43.0
```
does not include a `-E` option...

it also reads

```
Examples:
  Export input SVG (in.svg) to PDF (out.pdf) format:
	inkscape --export-filename=out.pdf in.svg
  Export input files (in1.svg, in2.svg) to PNG format keeping original name (in1.png, in2.png):
	inkscape --export-type=png in1.svg in2.svg
```

so maybe the long-form options `--export-filename` or `--export-type` are better options to keep compatibility?

PR #464, opened 2020-07-10
[faust2plot/faust2csvplot] allow specification of include directories

LFSaw:
this PR adds (crude) functionality of specifying additional search paths for `.lib` files to `faust2plot` and `faust2csvplot`.

### DIFF
--- a/tools/faust2appls/faust2csvplot
+++ b/tools/faust2appls/faust2csvplot
@@ -15,6 +15,7 @@ CXXFLAGS=$MYGCCFLAGS
 
 ARCHFILE=$FAUSTARCH/csvplot.cpp
 FAUSTFLOAT=float
+FAUST_INCLUDES="0"
 SOUNDFILELIBS=""
 
 #-------------------------------------------------------------------
@@ -39,7 +40,11 @@ for p in $@; do
         exit
     fi
  
-    if [ $p = "-arch32" ]; then
+    if [ $FAUST_INCLUDES = "1" ]; then
+        OPTIONS="$OPTIONS -I $p"
+        CXXFLAGS="$CXXFLAGS -I $p"
+        FAUST_INCLUDES="0"
+    elif [ $p = "-arch32" ]; then
         PROCARCH="-m32 -L/usr/lib32"
     elif [ $p = "-arch64" ]; then
         PROCARCH="-m64"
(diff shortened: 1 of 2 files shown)


sletz:
-I is supposed to be used for Faust libraries. I understand the PR wants to use it to give paths to the C++ compiler? What is the use-case in the first place?

LFSaw:
> -I is supposed to be used for Faust libraries. I understand the PR wants to use it to give paths to the C++ compiler? What is the use-case in the first place?

The PR adds functionality to use `-I` for the faust binary as well as for the C++ compiler. A Use-case for this is, if the libraries used (most probably custom-written) include `ffunction` definitions for functions declared in a custom `.h` header file. 

### Example use-case

library file in `/custom/lib/path`

```faust
// library 'bitNOT.lib'

bitNOT = ffunction(int bit_not (int), "bitNOT.h", "");
```

header file (since custom, non-standard header file, it needs to reside, e.g. in the same directory as the library file, i.e. `/custom/lib/path`)

```cpp
// header file "bitNOT.h"

int bit_not(int x) {
	return ~x;
}
```
dsp file

```faust
// myProg.dsp

process = float(bitNOT(_));
```

the call to faust2csvplot would then be

```bash
faust2csv -I'/custom/lib/path' myProg.dsp
```

sletz:
Adding the relevant -I /custom/lib/path in CXXFLAGS is the more standard way to do. I've fixed all relevant faust2xx scripts to behave like this, see https://github.com/grame-cncm/faust/commit/6cd834496f2743651350723324fcbd1d62a9f3a6.

LFSaw:
thanks for the effort to go through all those tools in 6cd834496f2743651350723324fcbd1d62a9f3a6. Much appreciated!

However, AFAICS this does not help to forward the `-I` option to the _faust_ compiler, or am I wrong? 


And a little comment: As a (possibly naïve) user of these tools, I somehow would expect a command-line option to add header files (/directories) (especially if I am not familiar with C++ development, which these tools are targeted towards)...

sletz:
> thanks for the effort to go through all those tools in [6cd8344](https://github.com/grame-cncm/faust/commit/6cd834496f2743651350723324fcbd1d62a9f3a6). Much appreciated!
> 
> However, AFAICS this does not help to forward the `-I` option to the _faust_ compiler, or am I wrong?

- I /path/ options are already given to the _faust_ compiler, like all options which are not specific to the script.
> 
> And a little comment: As a (possibly naïve) user of these tools, I somehow would expect a command-line option to add header files (/directories) (especially if I am not familiar with C++ development, which these tools are targeted towards)...

Yes, but using the -I /path for the Faust compiler and the underlying C++ seems like strange. And I any case if you know what -I /path/ has to be given to the C++ compiled to make the program compiler, then I guess using the CXXFLAGS variable is not out of reach. I've improved the documentation a bit here: https://github.com/grame-cncm/faust/tree/master-dev/tools/faust2appls

PR #466, opened 2020-07-21
[WIP] Improved rust autovectorization

bluenote10:
@sletz I think I need some help with this to get it finalized ;). I have found a basic way to make the code gen produce an iterator based for loop, but there is still some missing functionality, and your feedback would be highly welcome.

For basic DSPs the code gen already works, and produces a `compute` method like this now (`copy2` example):

```rust
fn compute(&mut self, count: i32, inputs: &[&[Self::T]], outputs: &mut[&mut[Self::T]]) {
    let (inputs0, inputs1) = if let [inputs0, inputs1, ..] = inputs {
        let inputs0 = inputs0[..count as usize].iter();
        let inputs1 = inputs1[..count as usize].iter();
        (inputs0, inputs1)
    } else {
        panic!("wrong number of inputs");
    };
    let (outputs0, outputs1) = if let [outputs0, outputs1, ..] = outputs {
        let outputs0 = outputs0[..count as usize].iter_mut();
        let outputs1 = outputs1[..count as usize].iter_mut();
        (outputs0, outputs1)
    } else {
        panic!("wrong number of outputs");
    };
    let zipped_iterators = inputs0.zip(inputs1).zip(outputs0).zip(outputs1);
    for (((input0, input1), output0), output1) in zipped_iterators {
        *output0 = (*input0 as f32);
        *output1 = (*input1 as f32);
    }
}
```

Performance looks pretty good with these changes (MB/sec output):

|           |   Rust |   Rust (iterators) |   C++ (no fastmath) |   C++ (fastmath) |
|:----------|-------:|-------------------:|--------------------:|-----------------:|
| copy1     | 5052.4 |            34314.1 |             19449.7 |          19261.8 |
| copy2     | 5040.1 |            40644.8 |             65471.4 |          65031.8 |
| math      |  951.5 |             6537.1 |              6688.1 |           7156.8 |

The first tricky part was to generate the pattern matching based destructuring of the input/output buffers. Technically the destructuring based approach is only strictly needed for the mutable buffer part (because the borrow checker then knows that no two different indexes into `outputs` are mutable borrowed at the same time). But the approach also makes sense on input side, because it gives a nice consistent error handling in case of insufficient number of channels. At first I tried to "fake" the whole expression with existing instructions, but that became incredibly difficult, and doesn't appear to be the right way. Since Rust simply has instructions that differ from the C++ oriented instruction set, I thought it makes the most sense to introduce a special instruction for that purpose, called `DeclareBufferIteratorsRust`, which can handle both the input and the output expression.

Similarly it probably does not make sense to model the concept of a "zipped iterator" with some abuse of existing iterators, so I decided to create a special `IteratorForLoopInst` for that purpose as well.

Open questions are:
- I don't understand how I should implement the for loop in `RustCodeContainer::produceInternal`. Somehow it uses the same `fCurLoop->generateSimpleScalarLoop()` as is used for the normal `compute`, but since it operates on different inputs/outputs (I assume), I'm not quite sure how I can make that work.
- Since I'm just producing an empty loop there now, some impulse tests don't work. Currently the `comb_delay1` impulse tests seems to run into an infinite loop in the Faust compiler, but maybe that is just a consequence of that.
- I wasn't confident enough to fully replace the other Rust specific `SimpleForLoopInst` for now, which is still used in `generateInitArray`, `generateShiftArray`, and `generateCopyArray`. Maybe this could also be turned into iterator based loops at a later point (if beneficial at all), so perhaps it makes sense to keep `SimpleForLoopInst` for the time being. Are there any simple DSP examples that produce these three loops?

### DIFF
--- a/compiler/generator/instructions.hh
+++ b/compiler/generator/instructions.hh
@@ -53,6 +53,7 @@ struct Printable;
 struct NullValueInst;
 struct NullStatementInst;
 struct DeclareVarInst;
+struct DeclareBufferIteratorsRust;
 struct DeclareFunInst;
 struct DeclareStructTypeInst;
 struct LoadVarInst;
@@ -190,6 +191,7 @@ struct InstVisitor : public virtual Garbageable {
     virtual void visit(DeclareVarInst* inst) {}
     virtual void visit(DeclareFunInst* inst) {}
     virtual void visit(DeclareStructTypeInst* inst) {}
+    virtual void visit(DeclareBufferIteratorsRust* inst) {}
 
     // Memory
     virtual void visit(LoadVarInst* inst) {}
@@ -260,6 +262,7 @@ struct CloneVisitor : public virtual Garbageable {
     virtual StatementInst* visit(DeclareVarInst* inst)        = 0;
     virtual StatementInst* visit(DeclareFunInst* inst)        = 0;
     virtual StatementInst* visit(DeclareStructTypeInst* inst) = 0;
+    virtual StatementInst* visit(DeclareBufferIteratorsRust* inst) = 0;
 
     // Memory
     virtual ValueInst*     visit(LoadVarInst* inst)        = 0;
@@ -494,7 +497,7 @@ struct StructTyped : public Typed {
         }
         return size;
     }
-    
+
     int getOffset(int field) const
     {
         int offset = 0;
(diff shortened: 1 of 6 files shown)


sletz:
This code was actually already merged in rust-wip, so now master-dev right ?

bluenote10:
Yes that would be great, thanks!

From my testing this was working well, hand the autovectorization can help a lot in terms of performance for the simple DSPs.

Issue #78, opened 2017-08-14
Jack-0.125.0 doesn't have 'jack_midi_reset_buffer'


yurivict:
I had to patch like this 's|jack_midi_reset_buffer|jack_midi_clear_buffer|' to make it build.
This function is either only in the older, or only in the newer versions of Jack.

Please correct!

sletz:
This function is in jack2 starting from 1.19.11.

yurivict:
1.9.11 is RC1 of Jack2.
The current versions are Jack1 0.125.0, and Jack2 1.9.10.
You need to add ifdefs to make sure it is available.

sletz:
Prepare a PR , thanks.

PR #478, opened 2020-08-13
faust2sndfile: add options to LINPUTS call, fix jack_midi_clear_buffer issue

LFSaw:
allows to run `faust2sndfile` with additional custom libraries:

```
faust2sndfile -I ../lib myDSP.dsp
```

### DIFF
--- a/architecture/faust/gui/ControlUI.h
+++ b/architecture/faust/gui/ControlUI.h
@@ -89,7 +89,7 @@ class ControlUI : public UI {
         void encodeMidiControl(void* midi_control_buffer, unsigned int frames)
         { 
             assert(fControlOut.size() <= frames);
-            jack_midi_reset_buffer(midi_control_buffer);
+            jack_midi_clear_buffer(midi_control_buffer);
           
             for (unsigned int i = 0; i < fControlOut.size(); i++) {
                 jack_midi_data_t* buffer = jack_midi_event_reserve(midi_control_buffer, i, 4);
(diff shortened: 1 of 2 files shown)


LFSaw:
related to https://github.com/grame-cncm/faust/issues/78

sletz:
Thanks. Partly used in https://github.com/grame-cncm/faust/commit/c236d23526de80c410f8014d36bb803af1da18b8.

LFSaw:
could you please elaborate if your commit fixes the issue i tried to fix?

sletz:
I've just copied the `LINPUTS=$(faust $OPTIONS -uim "$f" | grep FAUST_INPUTS)` line in faust2sndfile.

LFSaw:
Do I assume right that the issue of `jack_midi_reset_buffer` vs `jack_midi_clear_buffer` I fixed is related to jack1 vs jack2?

Also (and I know this is minor here but still want to ask) I am curious if there is any policy from grame that prevents you from modifying my commit and merging the PR into the master branch, preserving mixed authorship in the git version tree?

sletz:
- yes, the is a JACK related issue that we will not merge (since the fix is not correct anyway)
- well 

> Do I assume right that the issue of `jack_midi_reset_buffer` vs `jack_midi_clear_buffer` I fixed is related to jack1 vs jack2?

- yes, the is a JACK related issue that we will not merge (since the fix is not correct anyway)

> 
> Also (and I know this is minor here but still want to ask) I am curious if there is any policy from grame that prevents you from modifying my commit and merging the PR into the master branch, preserving mixed authorship in the git version tree?

- well the fix was quite trivial. In this case I usually add the author name in the commit, I'v just forgotten this case... sorry for that.

LFSaw:
> yes, the is a JACK related issue that we will not merge (since the fix is not correct anyway)

I'd like to use `faust2sndfile` on OSX but unfortunately it is relying on jack(2) (out of whatever reason...); this made the script compile for the jack version available through homebrew; how could I make this work for OSX (without compiling jack2 myself)?

> well the fix was quite trivial. In this case I usually add the author name in the commit, I'v just forgotten in this case... sorry for that.

nevermind 👍

sletz:
Well ControlUI.h header is not needed in the first place AFAICS, fixed in: https://github.com/grame-cncm/faust/commit/cc637fc301d0b74bd3a1325f6cf40b8c461e817f

Can you test and report ? thanks.

PR #483, opened 2020-08-27
added es8388 driver support

droosenb:
This adds support for the es8388 audio codec, the codec used by the espressif LyraT to faust.

-added three driver files, as well as the relevant makefiles to the faust architecture. They configure the es8388 codec based on input from the user and behave similarly to the wm8978 and ac101 drivers. 
-updated the faust2esp32 to include a -es8388 option, which behaves the same as the options for the other drivers while substituting the relevant es8388 files
-added support for the LyraT to esp32-dsp.h
-fixed a small typo wm8798 -> wm8978 in faust2esp32

the driver files are fully functional but still under development

### DIFF
--- a/architecture/esp32/drivers/es8388/CMakeLists.txt
+++ b/architecture/esp32/drivers/es8388/CMakeLists.txt
@@ -1,0 +1,5 @@
+idf_component_register(SRCS ./es8388.cpp
+                       INCLUDE_DIRS "include"                       # Edit following two lines to set component requirements (see docs)
+                       REQUIRES 
+                       PRIV_REQUIRES )
+
--- a/architecture/esp32/drivers/es8388/adf_structs.h
+++ b/architecture/esp32/drivers/es8388/adf_structs.h
@@ -1,0 +1,320 @@
+/*
+ * ESPRESSIF MIT License
+ *
+ * Copyright (c) 2019 <ESPRESSIF SYSTEMS (SHANGHAI) CO., LTD>
+ *
+ * Permission is hereby granted for use on all ESPRESSIF SYSTEMS products, in which case,
+ * it is free of charge, to any person obtaining a copy of this software and associated
+ * documentation files (the "Software"), to deal in the Software without restriction, including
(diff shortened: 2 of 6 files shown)


sletz:
@rmichon I understand you've developed something similar yes? Should we still merge this PR ?

rmichon:
This is WAY more sophisticated that what I wrote (which is more for pedagogical purposes than for something really practical). My driver is much lower-level and lightweight than this one. I think Dirk's version is the one that should make it to the Faust repo so I just merged his request. Thanks Dirk for putting this together!

cyberic99:
@droosenb can you elaborate a bit on the performance of such hardware?

what kind of FAUST code can you run on it?

thanks

droosenb:
@cyberic99 The LyraT has very similar performance capabilities to other esp32 development boards. 
Memory is a big constraint, limiting both wavetable size and delay line length, though less so than other embedded devices like the teensy. Processing power is limited but still capable. I did some stress testing this winter and had about 50 sawtooth oscillators running in parallel before it topped out. I'm happy to talk if you want to know more, email me droosenb [at] oberlin [dot] edu

mrSilkie:
https://faust.grame.fr/doc/tutorials/#dsp-on-the-esp32-with-faust

Tutorial states that the faust2esp32 won't work with this codec. I'm guessing this is no longer the case for newer lyraT models now that we have this es8388 support.

Should probably get updated, but just to clarify.
" faust2esp32 -lib FaustSawtooth.dsp "

Can now support es8388 codecs via;
faust2esp32 -lib -es8388 FaustSawtooth.dsp

droosenb:
Hi mrSilkie!
I've replied to your issue that you've opened. Hopefully that covers most
of your questions. If you have any other questions, go ahead and continue
the conversation there. You're also welcome to contact me via email.

rmichon:
@mrSilkie we're about to make a pretty large update to the Faust website. Not sure how long this will take exactly (a couple of weeks) but we'll make sure that this is fixed in it.

PR #485, opened 2020-08-28
MSVC compiler bug

agraef:
The bug causes the little state machine in evalLabel (eval.cpp) to "stutter" in some cases (e.g., `%k` becomes `%kk`). Apparently that's because in Release mode the compiler (MSVC from Visual Studio 2019, i.e., the latest one) optimizes the loop so aggressively that it misses out the increment of the `src` pointer when switching to state 2 after the first IdentChar is detected.

We work around this by just dropping the local `char c` variable and using the `src` pointer directly throughout the loop, incrementing it whenever necessary. This solves the issue.

NOTE: This PR is relative to the 2.27.1 release (actually 2.27.2, which is what I'm using as a submodule right now while working on pd-faustgen), but the changeset applies cleanly to master-dev as well. I can rebase it on master-dev if you want, just let me know.

### DIFF
--- a/compiler/evaluate/eval.cpp
+++ b/compiler/evaluate/eval.cpp
@@ -849,47 +849,44 @@ static string evalLabel(const char* src, Tree visited, Tree localValEnv)
     string format;     // current format
 
     while (state != -1) {
-        char c = *src++;
-
         if (state == 0) {
-            if (c == 0) {
+            if (*src == 0) {
                 state = -1;
-            } else if (c == '%') {
+            } else if (*src == '%') {
                 ident  = "";
                 format = "";
                 state  = 1;
+                src++;
             } else {
-                dst += c;
+                dst += *src++;
                 state = 0;
             }
 
         } else if (state == 1) {
-            if (c == 0) {
+            if (*src == 0) {
                 // fin et pas d'indentifiant, abandon
                 dst += '%';
                 dst += format;
                 state = -1;
-            } else if (isDigitChar(c)) {
-                format += c;
+            } else if (isDigitChar(*src)) {
+                format += *src++;
                 state = 1;
-            } else if (isIdentChar(c)) {
-                ident += c;
+            } else if (isIdentChar(*src)) {
+                ident += *src++;
                 state = 2;
             } else {
                 // caractere de ponctuation et pas d'indentifiant, abandon
                 dst += '%';
                 dst += format;
-                src--;
                 state = 0;
             }
 
         } else if (state == 2) {
-            if (isIdentChar(c)) {
-                ident += c;
+            if (isIdentChar(*src)) {
+                ident += *src++;
                 state = 2;
             } else {
                 writeIdentValue(dst, format, ident, visited, localValEnv);
-                src--;
                 state = 0;
             }
 


agraef:
I have no idea why https://ci.appveyor.com/project/orlarey/faust/builds/34910020 fails on this because the code contains no MSVC-centric changes and in fact the build look looks just fine (apart from the usual barrage of warnings MSVC gives when compiling Faust). I've tested this change myself on both Linux and Windows (64 bit), and it works for me.

NB: Please keep my msvc-compiler-bug branch around for the time being, as I'm using this as a submodule in pd-faustgen2. Thanks. :)

sletz:
" I can rebase it on master-dev if you want, just let me know." => yes please do !

agraef:
> " I can rebase it on master-dev if you want, just let me know." => yes please do !

Done, cf. #486. I created a new PR and so will close this one now, because I need the msvc-compiler-bug branch to hang around in order to build pd-faustgen2 with it.

BTW, Stéphane, do you know whether anyone has used the *32 bit* X86 LLVM backend successfully with Faust's JIT yet? I'm asking because it keeps crashing for me. Compiling the dsp works fine, but my Pd external crashes as soon as it tries to call the LLVM-generated code, which happens to be llvm_dsp::instanceConstants() being called from llvm_dsp::init(). This is with LLVM 9.0.0.

Everything works fine in 64 bit (x86_64), so I suspect that it's an ABI issue. But I already tried every target triple I could think of (it should be "i386-pc-win32", but I also tried various other combinations, and even different ABIs like pe32, gnu, and elf), all to no avail, it just keeps crashing in 32 bit land. Which is a bummer, since it means that on Windows faustgen2~ won't run in the 32 bit version of Pd, nor in Purr Data (which is 32 bit only right now).

Should I report this as a bug? I don't have a minimal example as a witness, just the external.

agraef:
NB: I still need the msvc-compiler-bug branch, so please don't delete it, thanks! :)

kmatheussen:
On Sat, Aug 29, 2020 at 5:29 PM Albert Graef <notifications@github.com>
wrote:

> " I can rebase it on master-dev if you want, just let me know." => yes
> please do !
>
> Done, cf. #486 <https://github.com/grame-cncm/faust/pull/486>. I created
> a new PR and so will close this one now, because I need the
> msvc-compiler-bug branch to hang around in order to build pd-faustgen2 with
> it.
>
> BTW, Stéphane, do you know whether anyone has used the *32 bit* X86 LLVM
> backend successfully with Faust's JIT yet? I'm asking because it keeps
> crashing for me. Compiling the dsp works fine, but my Pd external crashes
> as soon as it tries to call the LLVM-generated code, which happens to be
> llvm_dsp::instanceConstants() being called from llvm_dsp::init(). This is
> with LLVM 9.0.0.
>

Hi Albert,

I don't remember the details, but I was not able to make the 32 bit LLVM
backend work either. This was for the Faust instrument in Radium. It just
crashed when trying to run the code if I remember correctly. Could be a
different problem though.

agraef:
Hi Kjetil, thanks a lot, that's really valuable information! Sounds exactly like the problem I have -- in 32 bit, the plugin crashes as soon as it tries to execute any code generated by the LLVM JIT. This might well be an LLVM bug. I'm afraid that, at least as far as the MCJIT is concerned, nobody really cares much about 32 bit on anything anymore, not even Windows where lots of 32 bit code is still being used.

agraef:
On second thought, the issue might still be in Faust's LLVM backend. The Win32 ABI has some differences to 32 bit Linux and macOS systems. But I'd assume that someone would have checked by now that the generated LLVM IR is correct on Win32. Stéphane, do you have any further insights on this?

sletz:
No more insights and never tested myself.

agraef:
Mac guy through and through. ;-) Well then, we'll just have to live with it for now, I guess. Another item for the infinite TODO list...

sletz:
I guess the right way would be to move Pure Data to 64 bits now...

agraef:
> I guess the right way would be to move Pure Data to 64 bits now...

No kidding. :) Yep, that thought already crossed my mind, but it's not exactly a walk in the park. Purr's codebase was forked from Pd-l2ork, which in turn was forked from Pd-extended, which was based on a very old Pd version, and at the time vanilla Pd had no 64 bit support on Windows whatsoever. While we backported most of vanilla's new features and bugfixes over the years, Windows 64 bit cleanliness wasn't a priority so far, so there are still quite a few bumps on that road.

Anyway, you're right, that's clearly the way to go. In fact, after a day of work I have a "light" Purr build (vanilla-like, without most of the externals we normally ship with Purr) working in 64 bit now, and that might be good enough for my courses in the winter semester.

PR #487, opened 2020-09-06
@(0) is valid

mxmxyz:
`@` can delay its input by 0 samples, so saying that it "must be positive" could be confusing (especially since the example immediately following this paragraph uses a slider that starts at 0!)

"cannot be negative" seems more precise to me.


sletz:
Thanks. Fix committed in the faustdoc separated project: https://github.com/grame-cncm/faustdoc/commit/1752f02ffb16904666208aeef26bf92c86ccbf31.

PR #505, opened 2020-10-05
fix ir2dsp and add poly ir2dsp

MacroMachines:
fixed the ir2dsp python script so that it scales the values correctly and creates a correct sound upon testing / comparison with audio file.  Created the initial version of a poly ir2dsp script that scales all the frequency values to ratios and allows playing everything with midi.


rmichon:
Pierre-Amaury Grumiaux was the original contributor of ir2dsp but all these modifications seem correct to me. Thanks @MacroMachines !

PR #511, opened 2020-11-06
Teensy 4.1 PSRAM support

modularev:
This adds the `-ext` option to the faust2teensy applet which makes use of the external memory, soldered to the back of the Teensy 4.1. Opposed to the internal RAM2 section (which is limited to 512Kb) you can get up to 16MB with 2x ESP-PSRAM64H. This allows for much larger and more buffers, however also increases CPU usage. Still quite a usable option to be included in my opinion.

### DIFF
--- a/architecture/teensy/teensy.cpp
+++ b/architecture/teensy/teensy.cpp
@@ -83,8 +83,8 @@
 
 /*******************BEGIN ARCHITECTURE SECTION (part 2/2)***************/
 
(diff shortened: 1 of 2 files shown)


sletz:
Thanks, but as explained on Slack I would prefer a more generic "custom memory manager" solution, to be also used on other embedded platforms. I'll need to work on that ASAP. This patch is too hard-coded to be merged.

PR #512, opened 2020-11-08
Incoming blocks can be NULL

modularev:
treat input as silent when receiveReadOnly(channel) returns NULL as described in [documentation](https://github.com/PaulStoffregen/Audio/raw/master/new_objects.md)


sletz:
Thanks for the PR. 
I finally rewrote it a bit differently, see https://github.com/grame-cncm/faust/commit/cbfe6066951193c81c6ccfd20b7887fa0e0fe0fd. Can you teste and report?

PR #517, opened 2020-11-17
D backend for the compiler

abaga129:
## What's in this PR

- New option for the `-lang` compiler switch to generate D code.  `-lang dlang`
- Architecture file `minimal.d` that only contains the necessary classes to compile and access the UI controls.

## Reason
[Dplug](https://github.com/AuburnSounds/dplug.git) is a growing plug-in framework written entirely in D.  Dplug is the framework I develop with and I wanted access to the vast library of Faust DSP algorithms in it.  

## Notes
The generated `.d` file will not compile on it's own as it is dependent on `dplug:core` which contains a few methods for allocating objects and arrays.  It would probably be possible to recreate those methods in the `minimal.d` architecture file or have the compiler generate them.  If this is an issue I'll be happy to look into it.


sletz:
Thanks for the PR. But before accepting it, you'll have to add whet is needed for proper backend testing, as explained on discord channel:

We have testing tools for backends. This is https://github.com/grame-cncm/faust/tree/master-dev/tests/impulse-tests

There is a set of DSP reference files + reference outputs (as text files with the sample values) + machinery to compare. You'll have to add a target for the D backend.

abaga129:
Okay great, I will work on adding the necessary tests.

abaga129:
All of the tests are now compiling and running, so that's a big step.  About 1/3 of the tests are not passing.  I think the issue is probably in how I'm running the tests rather than the code being generated.

sletz:
Much better. Almost all tests are passing here. But:
- in your https://github.com/ctrecordings/faust/blob/master-dev/tests/impulse-tests/archs/architecture.d file and in `runDSP` function, you should really follow what is done in the `runDSP` function located in https://github.com/ctrecordings/faust/blob/master-dev/tests/impulse-tests/archs/controlTools.h especially concerning bufferSize management. One of the test` bs.dsp` fails because of this reason. I suggest trying following the C++ more precisely, using the same variable names for instance, otherwise this more difficult to understand was happens
- `math.dsp` fails probably because of a difference a math function, we'll have to dig that
- then you'll have to also test the` -vec` mode

abaga129:
Thanks for the suggestions.  I've rewritten the D architecture file for the tests to more closely follow the C++ implementation.  It's now down to just 6 failing tests.  I've commented out the SoundFile and MIDI parts of the test.  That still needs to be implemented somewhere.

The remaining failing tests are:

- [ ] karplus
- [ ] karplus32
- [ ] reverb_designer
- [ ] reverb_tester
- [ ] sound
- [ ] UITester

sletz:
Looking rapidly: this line is still incorrect:
https://github.com/ctrecordings/faust/blob/master-dev/tests/impulse-tests/archs/architecture.d#L313

`finterface.setButtons(true)` is used to simulate pressing buttons in the UI, you'll have to implement that (and also `finterface.setButtons(false);` of course)

This probably explains the failing tests.

abaga129:
All of the D backend tests should now be passing except for `sound.dsp`.  I havent implemented the Soundfile related portions of faust for D.  I see that the rust backend ignores this as well.  If you would prefer to have it though, I can try to implement it.

I havent tested with the `-vec` option yet.  The minimal architecture file for D will need some changes now so that will be my next step.  Other than that I'm not sure what else is left.

sletz:
Testing `-vec` is more important. Soundfile support probably less and would require more work.

abaga129:
All tests are passing and I've managed to use it in Dplug to create a simple reverb plug-in.  Let me know if there is anything else that needs to be done!

sletz:
Thanks. 
I see that `FAUSTFLOAT` is not used anymore as the "external" type. Is there any reason for that? The thing is that the  internal computation type (when `-float` or `-double` is used) can possibly be different from the external `FAUSTFLOAT` type use is UI class and compute for instance.  Thats is a `FAUSTFLOAT = float ` type can possibly be using with the `-double ` option (thus internal type = double).
Il would be preferable to keep the internal/external type distinction as done in the C/C++ backends.

sletz:
Another points:
- In dlang_instructions.hh why is `gPolyMathLibTable["remainder"] = "remainder", gPolyMathLibTable["rint"] = "round";` added in the `double` case and not `float` case ?
- Ifnot needed at  the end,  can you clean commented code ? (like `visit(Select2Inst* inst)` and  `visit(ForLoopInst* inst)` ? 

Thanks.

abaga129:
I see your point.  I'll work on cleaning up the code in the suggested areas.

I think I hastily added `gPolyMathLibTable["rint"] = "round";` to the end of the list under the `double` case when infact it inst related to `float` or `double`.

I'll also rework the architecture files to not use `FAUSTFLOAT`.  This makes a lot more sense to me.

sletz:
Hum... ((-; I was trying to explain exactly the contrary ! The architecture file **has to use** `FAUSTFLOAT` , and `FAUSTFLOAT` will be `double` or `float` depending go the final choice of the audio layer, and the generated code has to use `FAUSTFLOAT` in the compute method. 

Moreover this new `generateDSPInterface` method does not make real sense: we try to minimize what the compiler generates, and report everything else at the architecture file level.
Since the `dsp` base class can part of the architecture file, please keep it at this level.

In general the more you conform to what the C+ backend already does (for instance since it is the one the more similar the D one), the best it is.

The point for us it to minimize the difference between backends as much as we can, so that to decrease the maintenance cost of all those different backends.

abaga129:
> and the generated code has to use FAUSTFLOAT in the compute method.

Ah I understand now.  This is what threw me off I think.  In my latest commit, I've reverted those changes and I think it should be correct now!

I also renamed `architecture/minimal.d`->``architecture/minimal-dplug.d` since it just provides a way to get access to the parameters from the UI and is what I'm using for testing in Dplug.  I've created a new file called `architecture/minimal.d` which is based off of the minimal C example.

sletz:
I did a PR to fix some stuff here: https://github.com/ctrecordings/faust/pull/4

abaga129:
Thanks @sletz, I've merged your PR.  I see now that I forgot to move the `dsp` interface back to the architecture files and misunderstood how the `compute` prototype is supposed to be implemented.

abaga129:
@sletz I've added the script!  It's fairly basic but has options for the important things in Dplug such as plugin id and vendor name.  Let me know if there is anything else necessary before this can be merged.  There is certainly a lot more work that could be done for the D architecture file: UI and MIDI are two that would be nice to have.

sletz:
Thanks !
 I've manually rebased your branch, changed the version number and merged everything in https://github.com/grame-cncm/faust/commit/1dcaa9159c03ad59aa56fa24c712eaa24e0d7f2a. 
Thanks a lot for the contribution. 
Please start gain from a fresh master-dev branch if you need to create new PR.

PR #535, opened 2020-12-21
Support option -ns with D language

jpcima:
This implements the flag `-ns` in dlang.
The behavior is that if the option is `-ns foo`, then the module is `foo.mydsp`.

This matches to a file named `foo/mydsp.d` in D's search path for imports; since D's file paths and module naming have to match, having this option is important.

It's slightly different from the bit of implementation present before, which wouldn't introduce the `.` character, but I thought it made more consistency with the cpp one.

Implements also the special case depending whether an architecture file is used or not.

### DIFF
--- a/compiler/README.md
+++ b/compiler/README.md
@@ -124,7 +124,7 @@ Code generation options:
 
   **-mapp**      **--math-approximation**         simpler/faster versions of 'floor/ceil/fmod/remainder' functions.
 
-  **-ns** \<name> **--namespace** \<name>           generate C++ code in a namespace \<name>.
+  **-ns** \<name> **--namespace** \<name>           generate C++ or D code in a namespace \<name>.
 
 
 Block diagram options:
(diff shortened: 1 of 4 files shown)


jpcima:
Under the case of architecture file though, it broke the `dub.sdl` meaningful comment which is desired to be at file's top.

There's possibly a smarter refactor to make such that direct C++ or D code needn't be output at libcode.cpp "enrobage" level, but I'm not totally sure which.
It should output first, and in order: `dub.sdl` comment, and `module` statement

sletz:
Note that `-ns ` option is not really working for C++. Should I merge this one for Dlang backend anyway? Is it ready enough?

jpcima:
Yes I suppose, if the solution e03ccce in libcode.cpp is considered fine enough.
Like the C++ one, this feels kinda hackish, but at least doesn't risk problems such as `#include` 's getting caught inside namespace imbrications.

This worked in all 4 combinations of presense or absence of `-a`/`-ns` flags.

sletz:
Thanks!

PR #551, opened 2021-02-16
quantizers example

CesarChaussinand:
I added an example on how to use the quantizers library.
@rmichon

### DIFF
--- a/examples/quantizing/quantizedChords.dsp
+++ b/examples/quantizing/quantizedChords.dsp
@@ -1,0 +1,47 @@
+/*
+MIT License
+
+Copyright (c) 2021 César Chaussinand
+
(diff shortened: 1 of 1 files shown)


sletz:
Thanks. I see that the generated code is quite costly. Using the FIR backend that gives statistics on the code,  I see:
` [atanf = 21 floorf = 4 logf = 3 max_f = 3 max_i = 3 min_f = 3 powf = 6]` operation of each sample in the DSP loop.

Could the code be optimised using techniques described here https://faustdoc.grame.fr/manual/optimizing/ ? possibly the `tabulate` approach ?

rmichon:
I just accepted the pull request but I do agree with Stéphane that some of the functions of the library could be optimized.

PR #569, opened 2021-04-03
Add header gards to JuceGUI.h

glocq:
Other header files feature guards, so I doubt the omission was voluntary.


sletz:
Unvoluntary, yes, thanks.

PR #576, opened 2021-04-19
Faust2juce fixes (mainly for Windows)

DBraun:
This PR updates `faust2juce` to Projucer 6.0.8 (build date 22 Mar 2021). As a result, some fields in the Projucer file get renamed, and some get added. Visual Studio 2013 was deprecated and removed. Visual Studio 2019 was added. The `faust2juce` script has a new optional argument `-vst2sdkdir` for a legacy VST 2 SDK folder, which enables the building of dll files as opposed to `.vst3` files. There's a new optional argument `-disable-splash-screen` for disabling JUCE's splash screen. By default, the splash screen is enabled. A bug with `grep "declare nvoices"` was fixed because the metadata might actually be `[nvoices:8]` which wouldn't be found by that grep. I tested on Windows by building a dsp with a process and and effect. The standalone exe, dll, and vst3 all work perfectly. **I have not tested on other platforms.**

### DIFF
--- a/architecture/juce/README.md
+++ b/architecture/juce/README.md
@@ -24,7 +24,7 @@ Two different achitecture files will be used to glue the previously described fi
 
 **faust2juce** is used with the following command: 
 
-`faust2juce [-osc] [-midi] [-soundfile] [-nvoices <num>] [-effect auto|<effect.dsp>] [-standalone] [-jucemodulesdir <dir>] [-jsynth] [-llvm] [additional Faust options (-vec -vs 8...)] file.dsp` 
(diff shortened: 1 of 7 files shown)


sletz:
Thanks. Testing here on macOS: in the previous version, a quite useful "standalone plugin" target was available (see screenshot). Can you restore it? Thanks.

Capture d’écran 2021-04-19 à 13 20 57

DBraun:
Sorry about that. I think this will fix it.

sletz:
Thanks!

Issue #585, opened 2021-05-23
Trouble creating VST3 via faust2juce with JUCE 6.0.8


Digital-Larry:
You're in luck, I discovered the fix.

Edit the .jucer file which is created by faust2juce (I exported from the online Faust IDE to juce->plugin).

In the \<JUCERPROJECT\> section:
#1 set buildVST="0", buildVST3="1"
#2 change "buildVST" in \<pluginFormats\> section to "buildVST3".

In the \<JUCEOPTIONS\> section:
#3 add JUCE_VST3_CAN_REPLACE_VST2="0".

sletz:
Can you they with the ones in master-dev branch ? And fix them with a Pull Request ?

Digital-Larry:
@sletz I probably would want to try this by building the code first as I don't understand all of the possibilities of people wanting to build VSTs which replace VST2 format DLLs.  That said, I did some web searching for Faust/VST and didn't find a lot, so if you're willing to abandon VST2 completely in favor of today's JUCE and VST3, I could make the changes.   I don't know how you handle CI/regression testing so I really don't want to break anything without being able to test it.

sletz:
Can you possibly test with a local version of Faust?

Digital-Larry:
I'm sure I can, you'll just need to be a little patient.  But I'll do it.

Digital-Larry:
Was resolved by https://github.com/grame-cncm/faust/pull/587

PR #587, opened 2021-05-30
Fix for issue #585, creating VST3 via JUCE

Digital-Larry:
See: https://github.com/grame-cncm/faust/issues/585

### DIFF
--- a/architecture/juce/plugin/plugin.jucer
+++ b/architecture/juce/plugin/plugin.jucer
@@ -177,7 +177,7 @@
     <MODULE id="juce_opengl" showAllCode="1" useLocalCopy="0"/>
     <MODULE id="juce_osc" showAllCode="1" useLocalCopy="0"/>
   </MODULES>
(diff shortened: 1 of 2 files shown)


Digital-Larry:
I use the value of BUILD_VST (which is set to 1 when you use -vst2sdkdir to set the path to the VST2 SDK folder) to also set the value of JUCE_VST3_CAN_REPLACE_VST2 in the generated jucer file.  I was trying to build VST3 only in JUCE, since the VST3 SDK is built-in, but it was failing while looking for aeffect.h and aeffectx.h (which I guess are part of the VST2 SDK).  As it is not even possible to get the VST2 SDK from Steinberg anymore, I was not able to test this in the case where you're trying to really build a VST2.  I figured, if you are trying to build a VST2 and a VST3 from the same code at the same time, I'm not entirely sure why you would do that, but in any case, if you are building VST2 and VST3 at the same time, and you DON'T want the VST3 to be able to replace the similarly named VST2, then you'll have to edit the jucer file manually to make that change.  I think that going forward most people are just going to make VST3s unless they have some special reason not to.

Issue #589, opened 2021-06-05
faust2juce sets IS_SYNTH=1 when "nvoices" was NOT found in the dsp file


Digital-Larry:
The issue is that "grep -c" is used to search for the pattern rather than simply "grep".  The return value of "grep -c" will be 1 if the pattern is NOT found, and 0 if the pattern IS found.

Here is the section of code involved:

#look for polyphonic "nvoices" metadata in the DSP file
if [ $(cat $FILES | grep -c "nvoices:") ] || [ $(cat $FILES | grep -c "declare nvoices") ]
then
        # a variation of the string "nvoices" was found
        IS_SYNTH="1"
 else
    # "nvoices" was not found
        IS_SYNTH="0"
fi

$ cat voice.dsp
nvoices:
$
$ grep -c ncoix voice.dsp
0
$ echo $?
1
$ grep -c nvoices voice.dsp
1
$ echo $?
0
$

Digital-Larry:
Attached is a test script to create a series of short dsp files exercising the various options.  Note that if "nvoices" is commented out, it will still fool the script as it's not set up to distinguish a commented line.

[test.zip](https://github.com/grame-cncm/faust/files/6602849/test.zip)

The output (of my proposed fix TBD) is:

$ ./test.sh
======= test file ==================
import("stdfaust.lib");
process = _,_;
======= synth related lines in created jucer file ==================
              pluginChannelConfigs="" pluginIsSynth="0" pluginWantsMidiIn="0"
======= test file ==================
import("stdfaust.lib");
declare options "[nvoices:5]";
process = _,_;
======= synth related lines in created jucer file ==================
              pluginChannelConfigs="" pluginIsSynth="1" pluginWantsMidiIn="0"
              pluginCharacteristicsValue="pluginIsSynth">
======= test file ==================
import("stdfaust.lib");
declare nvoices "5";
process = _,_;
======= synth related lines in created jucer file ==================
              pluginChannelConfigs="" pluginIsSynth="1" pluginWantsMidiIn="0"
              pluginCharacteristicsValue="pluginIsSynth">
$

Digital-Larry:
I updated the test.sh, and it works properly with -nvoices on the command line and not in the dsp file.

[test.zip](https://github.com/grame-cncm/faust/files/6602921/test.zip)

```
$ ./test.sh
======= test file ==================
import("stdfaust.lib");
process = _,_;
======= command line ==================
faust2juce  test.dsp
[...]

PR #590, opened 2021-06-05
Fix logic around setting "IS_SYNTH" in faust2juce

Digital-Larry:
Fixes https://github.com/grame-cncm/faust/issues/589


Digital-Larry:
Still not right.

PR #600, opened 2021-06-21
WM8978 codec change Sample Rate. 

andy-wi:
Additional functions to set the Sample Rate of the codec, enable/disable ADC to DAC loopback and enable/disable DAC automute as per datasheet.


sletz:
"This branch has conflicts that must be resolved" : can you fix that? Or prepare a new PR for the new commits?

PR #603, opened 2021-06-23
Additional functions for WM8978

andy-wi:
Additional functions for WM8978 codec; power saving, enable mono output, change sample rate, speaker boost, loopback and automute.


sletz:
Thanks !

PR #617, opened 2021-07-13
Improve integration with Foley's Plugin GUI Magic package

Digital-Larry:
See:  https://github.com/ffAudio/PluginGuiMagic
Use the "-magic" flag with faust2juce.
This should create a preprocessor definition "PLUGIN_MAGIC".  
Tested with PGM version 1.13.
 e.g. VisualStudio 2019

image

There are a couple of other options as well:

MAGIC_LOAD_BINARY 
- is supplied, default not defined, lets you "bake in" your XML file by adding it to your Jucer project and then defining this variable, either in the Jucer file or directly in the cpp file.  You also have to change the names of the variables magic_xml and magic_xmlSize in the code to match your file name.  If you don't you'll get a compilation error, so you'll know where to look.

MAGIC_LEVEL_SOURCE
- is supplied, default not defined, just to show you where in the source code you would add those sorts of things.  Knowledge of how to use the PGM components is assumed.


architecture/juce/juce-plugin.cpp
```diff
@@ -237,6 +242,9 @@ class FaustPlugInAudioProcessor : public foleys::MagicProcessor, private juce::T
     void processBlock (juce::AudioBuffer<float>& buffer, juce::MidiBuffer& midiMessages) override
     {
         jassert (! isUsingDoublePrecision());
+#ifdef MAGIC_LEVEL_SOURCE
+	outputMeter->pushSamples(buffer);
```
ffAudio, line 246: If you push the samples before the `process (buffer, midiMessages)` it is actually an inputMeter ;-)
You could have both though if you want.

architecture/juce/juce-plugin.cpp
```diff
@@ -627,6 +652,10 @@ void FaustPlugInAudioProcessor::prepareToPlay (double sampleRate, int samplesPer
     fDSP->instanceClear();
     
 #endif
+#ifdef MAGIC_LEVEL_SOURCE
+    magicState.prepareToPlay(sampleRate, samplesPerBlock);
+    outputMeter->setupSource(2, sampleRate, 500, 200);
```
ffAudio, line 657: Would be better to use `getMainBusNumInputChannels()` and `getMainBusNumOutputChannels()` for the output

architecture/juce/juce-plugin.cpp
Digital-Larry: Will review.

architecture/juce/juce-plugin.cpp
Digital-Larry: Will update.

architecture/juce/juce-plugin.cpp
Digital-Larry: Somehow I had deleted process (buffer, midiMessages) from this completely while I was working on the code.  It's pretty difficult to tell whether the metering is on input of output.  Where is there further documentation on this?

ffAudio: If you push the samples to the meter before processing it's input, if you push after processing, it's output. ;-)
If you click on "View changes" you will see the side-by-side diff, which shows a bit more context.

Digital-Larry: Ah, I guess it makes sense.  The buffer is changed after the process() call.

sletz:
I suggest adding a section in the [README here](https://github.com/grame-cncm/faust/tree/master-dev/architecture/juce) about the PluginGuiMagic integration, explaining the use of MAGIC_LOAD_BINARY and MAGIC_LEVEL_SOURCE and everything needed.

Digital-Larry:
@sletz I updated the readme.MD per your suggestion.

Digital-Larry:
I'm done making changes for the time being.

sletz:
Thanks. I will wait for official 2.33.1 release to merge this. This will simplify some merging issues.

sletz:
Merger in --squash mode here: https://github.com/grame-cncm/faust/commit/106a187d31b47e71ecdb7a0d16612845ebcd8302
Thanks.

PR #618, opened 2021-07-15
Annot faust

adud:
add interval annotation in Faust syntax and scalar ocpp compiler


sletz:
Can ou possibly look at merge conflicts?

PR #620, opened 2021-07-15
audioinput example

ArcAudio:
dunno if "audioinput-processor.js" is strictly necessary.


sletz:
I'm not sure: how is the example supposed to work? Where is the processing audio file? (edited)

Issue #631, opened 2021-08-03
faust2juce draws white on white gui by default


madskjeldgaard:
Hello 

This is a tiny issue but the faust2juce tool creates GUI's with white on white by default:
![2021-08-03-16:14:57-screenshot](https://user-images.githubusercontent.com/22474608/128031402-23efef9b-2f13-4aab-9863-34ec03bd097f.png)

I've fixed it for myself here by changing this line to fill using black instead:
https://github.com/grame-cncm/faust/blob/76ab7cbc48f9f20249ce19891ac8eab0cd5ec8cb/architecture/juce/juce-plugin.cpp#L770

That fixes the issue. I can do a PR for this if needed.

sletz:
Or better change the font colour from white to something else?

mouseos:
/usr/local/include/faust/gui/JuceGUI.h
and change "black" to "white" in
/usr/local/share/faust/juce
in /usr/local/include/faust/gui/JuceGUI.h to "white", and "g.fillAll (juce::Colours::white);" to "g.fillAll (juce::Colours::black);".
![Screenshot_20211215-233125](https://user-images.githubusercontent.com/46545607/146206730-5fede179-2af4-4114-820e-64e8f27bd3a9.png)

PR #632, opened 2021-08-04
faust2juce: Change background colour to black

madskjeldgaard:
Fixes https://github.com/grame-cncm/faust/issues/631

This is what it looked like before:
2021-08-03-16:14:57-screenshot

This is what it looks like now:
2021-08-04-10:00:15-screenshot

(sorry if this is a bit lazy - but I've got limited time for this ATM. Best!)


simonhillfr:
I'm running into the same issue, there is no contrast.
It looks like there is a `CustomLookAndFeel` class that has more contrast but ends up never being used. Should this class be removed ?
Changing the background colour gets the job done. I find that`darkgrey` looks less harsh than black

PR #641, opened 2021-08-24
Fast vhdl

bouksimo98:


sletz:
Merged here https://github.com/grame-cncm/faust/commits/fast-vhdl, thanks.

Issue #645, opened 2021-09-02
[coreaudio] coreaudio doesn't work as expected when the chosen input device has outputs


nuchi:
I'm trying to send input from [BlackHole](https://github.com/ExistentialAudio/BlackHole/) (16 inputs and 16 outputs) to my external headphones (2 outputs). The problem is that the aggregate device created from the input and output device has a combined 16 inputs but _18_ outputs, not the 2 I'd expect. The headphones' outputs are the _last_ two channels, but the coreaudio driver is trying to send the output to the _first_ two channels.

To reproduce:
`test.dsp`:
```
process = _,_;
```
and then:
```
% faust2api -coreaudio test.dsp -nozip
% clang++ DspFaust.cpp -DBUILD -std=c++11 -o DspFaust -framework CoreAudio -framework CoreServices -framework AudioUnit
% ./DspFaust
Type 'q' to quit
```
and playing sound to the BlackHole output produces nothing audible. Trying the same passthrough in AU Lab (going from BlackHole input to headphone output) works correctly.

Editing the following line in the generated `DspFaust.cpp` causes audio to be passed through as expected, seeming to confirm that the headphones' two outputs are being stuck after the input device's outputs:
```diff
  if (outChan < fPhysicalOutputs) {
      SInt32 chanArr[fPhysicalOutputs];
      for (int i = 0; i < fPhysicalOutputs; i++) {
          chanArr[i] = -1;
      }
      for (int i = 0; i < outChan; i++) {
-         chanArr[i] = i;
+         chanArr[i + 16] = i;
      }
      err = AudioUnitSetProperty(fAUHAL, kAudioOutputUnitProperty_ChannelMap, kAudioUnitScope_Output, 0, chanArr, sizeof(SInt32) * fPhysicalOutputs);
      if (err != noErr) {
          printf("Error calling AudioUnitSetProperty - kAudioOutputUnitProperty_ChannelMap 0\n");
          printError(err);
      }
  }
```
[...]

PR #647, opened 2021-09-04
[coreaudio] Skip an input device's output channels

nuchi:
Fixes https://github.com/grame-cncm/faust/issues/645

cc @sletz — I built my example locally (see the issue for my instructions to reproduce) and verified that this fix works, under a variety of input devices, both with and without output channels. I couldn't find any unit tests for anything in this file.

By the way I'm not sure what was wrong with your original approach. What I do here is add up all the output channels belonging to the all input subdevices, and then start the `kAudioOutputUnitProperty_ChannelMap` at that offset.

### DIFF
--- a/architecture/faust/audio/coreaudio-dsp.h
+++ b/architecture/faust/audio/coreaudio-dsp.h
@@ -235,7 +235,7 @@ class TCoreAudioRenderer
         AudioUnit fAUHAL;
         bool fState;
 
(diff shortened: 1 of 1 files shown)


sletz:
Merged here https://github.com/grame-cncm/faust/commit/6f1495472ed3beccd4be839c7a81a23068a008fc thanks!.

PR #679, opened 2021-12-06
Sublime Text 4 completions and snippets

Simon-L:
This PR proposes to add Sublime Text 4 automatic completions for the standard library and snippets taken from the atom snippets.

I'd like to have that tested a little bit before requesting the merge, this PR can be used for this. Is anyone using Sublime Text 4? @nuchi, maybe?

I wanted to bring as much as possible of the ease of use of faustide to my favorite editor. It resulted in doing more intensive parsing of the `.lib` files in order to extract Usage information, as well as a direct link to the docs:
faust-completion

I don't know about other code editors but Sublime Text is not very generous with the size of the area provided for completion: it has to be a single line and I couldn't figure how to force it to span according to the length of the content.
Some functions from the library have long usage lines and the worst case that I haven't covered is when the Usage has multiple lines and the first line is not the one demonstrating the usage. Only the first line of the Usage section is kept.

The more complex parsing allowed to spot some typos in the libraries, thanks to @sletz for promptly correcting all of these!

That is to say, this may have let a few incorrect names, links or usage information through, I cannot expect the automated script to catch every case, feel free to report any that you may find.

### DIFF
--- a/syntax-highlighting/README.md
+++ b/syntax-highlighting/README.md
@@ -80,3 +80,8 @@ Copy the directory `atom/language-faust` into `~/.atom/packages/` and follow the
 3. Click Install to automatically download and install the extension.
 
 NB: This is the latest version of [faust extension](https://marketplace.visualstudio.com/items?itemName=glen-anderson.vscode-faust) created by Glen Anderson, cf. https://github.com/hellbent/vscode-faust.
+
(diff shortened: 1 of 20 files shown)


nuchi:
Wow, this looks great!

PR #680, opened 2021-12-06
Override the default keymap for single quote (both insertion and deletion) so that it can be used for delay and letrec definitions

nuchi:
Sublime Text default behavior for single quotes is to pair them. This doesn't make sense when editing Faust files, where single quotes refer to delay (or a letrec definition) and not to strings.

This addition overrides the default behavior (for `source.faust` scope).

attn @sletz @Simon-L

### DIFF
--- a/syntax-highlighting/sublime-text/Faust/Default.sublime-keymap
+++ b/syntax-highlighting/sublime-text/Faust/Default.sublime-keymap
@@ -1,0 +1,18 @@
+[    // Don't auto-pair single quotes
+    { "keys": ["'"], "command": "insert_snippet", "args": {"contents": "'"}, "context":
+        [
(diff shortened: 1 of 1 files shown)


sletz:
@Simon-L OK for you ?

Simon-L:
Absolutely yes! Thanks @nuchi 
Also at that point would it make sense to have your repo (https://github.com/nuchi/faust-sublime-syntax) here now that we have a dedicated Sublime Text folder? :)
Note that I haven't tested if putting all the files in the same directory works as well, although I doubt it would work any different.

nuchi:
P.S. I've tested this by symlinking the `sublime-text/Faust` directory into `Packages/User` and everything works fine.

Issue #686, opened 2021-12-31
CSharp backend generating invalid code for "IsInfinity"


mikeoliphant:
At some point the "IsInfinity" operator was added, and the C# code generator produces invalid code.

it generates:

`public static bool IsInfinity(double d) { return Double.IsNegativeInfinity(double d) || Double.IsPositiveInfinity(double d); }`

but it should be:

`public static bool IsInfinity(double d) { return Double.IsNegativeInfinity(d) || Double.IsPositiveInfinity(d); }`

Also, this is inserted into the main body of the output file instead of inside the class, which is invalid.

Instead of being generated, the function should probably be put in the base class, **CSharpFaustBase.cs**, like _FMod_ and _FModF_ are.

sletz:
Can you fix it and prepare a PR ? Thanks.

Issue #687, opened 2021-12-31
GetInputRate and GetOutputRate need to be removed from CSharp interface


mikeoliphant:
It looks like the **GetInputRate** and **GetOutputRate** functions have been removed.

They also need to be removed from the _**IFaustDSP**_ interface in **CSharpFaustBase.cs**.

sletz:
Please prepare a PR, thanks.

PR #690, opened 2022-01-01
CSharp fixes for IsInfinity, removal of GetInputRate() and GetOutputRate()

mikeoliphant:
Added IsInfinity and IsInfinityF to C# base class and removed from codegen. Removed GetInputRate() and GetOutputRate() from IFaustDSP interface.

Fixes issues #686 and #687


sletz:
Thanks!

Issue #640, opened 2021-08-19
Web IDE - Browser support


suyashjoshi:
Please support Safari browser for Web IDE. Which browser does it currently run on, Chrome?

sletz:
Is runs on Chrome with MIDI support and FireFox. Both of then correctly support AudioWorklet.
Safari still has a [bug in AudioWorklet support](https://bugs.webkit.org/show_bug.cgi?id=220038), but can run in the older ScriptProcessor mode.

PR #697, opened 2022-01-26
Work around Safari bug for wasm support

nuchi:
Fixes https://github.com/grame-cncm/faust/issues/640

I've tested with `faust2wasm -worklet`, the result works across Firefox/Chromium/Safari on macOS. I haven't tested the other code paths but they should work just as well. I also haven't regenerated the web audio examples.

The Safari bug means that wasm modules can't be posted to the AudioWorketProcessor — the workaround is to post the buffer and compile in the processor.

### DIFF
--- a/architecture/webaudio/webaudio-workletnode-emcc-poly-standalone-wrapper.js
+++ b/architecture/webaudio/webaudio-workletnode-emcc-poly-standalone-wrapper.js
@@ -481,7 +481,7 @@ class mydspPoly {
 
             let json_object = JSON.parse(json);
 
-            let options = { wasm_module: wasm_module.module, json: json, wasm_effect_module: ((wasm_effect_module) ? wasm_effect_module.module : null), json_effect: json_effect, soundfiles: soundfiles, json_all: json_all, polyphony: this.polyphony };
+            let options = { wasm_buffer: wasm_buffer, json: json, wasm_effect_module: ((wasm_effect_module) ? wasm_effect_module.module : null), json_effect: json_effect, soundfiles: soundfiles, json_all: json_all, polyphony: this.polyphony };
 
             let re = /JSON_STR/g;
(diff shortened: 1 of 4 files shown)


sletz:
Thanks, but I will probably not apply this PR: the point of compiling on the node side is to share the compiled module and only do the (light...) instantiation on the processor side, which is especially important if several DSP instances (so  several processors) are instantiated.

The WebKit/Safari problem is identified and [worked here](https://bugs.webkit.org/show_bug.cgi?id=220038). I prefer waiting for a clean fix.

nuchi:
Ah, thanks for the explanation — probably there's still some stuff I don't understand, because I didn't realize there were any negative tradeoffs to this fix. Certainly some up-front CPU cost if each instance has to do some compilation instead of it happening once, but I figured that since that happens in the constructor then it's just a one-time cost and that that would be acceptable. But I admittedly haven't done any memory or CPU profiling, and don't know enough to know if there are any other costs to this approach.

sletz:
I would definitely prefer having Apple do the proper fix in  the first place. So I've [asked again](https://bugs.webkit.org/show_bug.cgi?id=220038).

ceelian:
As this is still an issue, might there be a way to have this PR or similar "workaround" merged until Apple is fixing the Safari bug? That would improve the developer experience a lot by not having to keep a locally modified version of FAUST just to support Safari users. Thx @nuchi for providing that fix/workaround.

sletz:
AFAICS Safari Preview now solve this issue, but still has a lot of audio glitches. Better put pression on Apple to improve the Web audio experience.

ceelian:
Thanks @sletz for the quick reply. Okay, until the Safari bug is fixed I forked the FAUST repo and applied the patch from @nuchi to the newest release V2.40.0. In case anyone else needs the workaround on that version: https://github.com/grame-cncm/faust.

PR #700, opened 2022-01-29
Add setTailLengthSeconds function

DBraun:
Suppose the real tail length of a polyphonic voice is only 0.1 seconds. The current default tail length of 0.5 seconds is too conservative, and it will lead to false alarm reports of voices being stolen: https://github.com/grame-cncm/faust/blob/5fcf858e0143b3c2a38257f17b7ac029836c46ed/architecture/faust/dsp/poly-dsp.h#L632 If the API supported setting the tail length, those false alarms could be avoided. I tested this by using createPolyDSPInstance, casting it to dsp_poly_effect with `dynamic_cast<dsp_poly_effect*>` and then `setTailLengthSeconds(0.01)` The audio output is the same but the warnings went away. It's a little unintuitive to cast to dsp_poly_effect, but that's what's returned by createPolyDSPInstance.

### DIFF
--- a/architecture/faust/dsp/poly-dsp.h
+++ b/architecture/faust/dsp/poly-dsp.h
@@ -167,6 +167,7 @@ struct dsp_voice : public MapUI, public decorator_dsp {
     int fDate;                          // KeyOn date
     int fRelease;                       // Current number of samples used in release mode to detect end of note
     FAUSTFLOAT fLevel;                  // Last audio block level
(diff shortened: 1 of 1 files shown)


sletz:
I've reworked the proposal a bit and committed [here](https://github.com/grame-cncm/faust/commit/29eba9218c9d78ef8bd5d8fd61db8c91f9e42e94). Thanks !

PR #704, opened 2022-02-09
fix: faust2sc.py: SuperNova plugins now compile with -DSUPERNOVA

madskjeldgaard:
fix: faust2sc.py: SuperNova plugins now compile with -DSUPERNOVA


madskjeldgaard:
Whoops - I had forgotten to add the `-DSUPERNOVA` flag so supernova objects actually weren't supernova objects. sorry

madskjeldgaard:
but this fixes it

PR #711, opened 2022-02-20
Update faust2caqt for QT6

olilarkin:
Without this change faust2caqt fails due to the usage of std::filesystem in QT6, with errors such as:

```/opt/homebrew/lib/QtCore.framework/Headers/qfileinfo.h:120:14: error: '~path' is unavailable: introduced in macOS 10.15
    { return QtPrivate::toFilesystemPath(absoluteFilePath()); }```

if supporting macOS < 10.15 is important, then this might not be the best fix

### DIFF
--- a/tools/faust2appls/faust2caqt
+++ b/tools/faust2appls/faust2caqt
@@ -134,8 +134,8 @@ if [[ $(uname) == Darwin ]]; then
     if [ $SYS_VERSION -ge 13 ]
 	then
 	    SPEC="-spec macx-clang"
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by olilarkin** [comment]
Actually, this is not enough.

QT6 has a dependency on librotli it seems and it's not in /usr/local/lib where the .app expects it.

It's possible to fix with 
```
brew install librotli
sudo ln -s /opt/homebrew/lib/libbrotlicommon.1.dylib /usr/local/lib/
```

but that means the .app is not portable

PR #713, opened 2022-02-25
Initial work for LLVM 15, also refactor the LLVM version macros

jcelerier:
Hi,
I tried building against LLVM 15 but it currently fails... 

I marked the places which changed and that I could not fix, pretty much only IRBuilder, but my LLVM-fu is not high enough to understand what a correct fix would be..

I also replaced the LLVM_XX macros by LLVM_VERSION_MAJOR checks (which have been there since llvm 3.1, from what I could see the earliest macros in the code refer to 3.4)


jcelerier:
yes, it's still at least 5/6 months away, I just sometimes test with llvm git head to be prepared in advance aha

PR #716, opened 2022-03-11
Repair polyphony with midi key-number only

modularev:


sletz:
Thanks, merged in https://github.com/grame-cncm/faust/commit/66ef64db07fb098fdc53582bd328d6d2bab3c27b

PR #721, opened 2022-04-06
Wasm2

dfober:
faust library for node


sletz:
[faustwasm](https://github.com/grame-cncm/faustwasm) is now the official project.

PR #724, opened 2022-04-10
disambiguate cmake target and command

pingdynasty:
I've been having problems building FAUST. For example:
```
$ make
cmake --build faustdir --config Release 
make: cmake: Permission denied

$ make clean
cmake --build faustdir --target clean
make: cmake: Permission denied
```
I can execute `cmake` from the command line with no problem.

I think the reason is that `cmake` in the Makefile is both a command and a target. Ideally I think the `cmake` target should be renamed, but I understand that changes to the build interface might have repercussions elsewhere.

Instead I'm submitting this patch where calls to `cmake` are done using the `$(CMAKE)` syntax. And `cmake` is added to `.PHONY`.

This however is not enough for the build to work for me, I also have to define `CMAKE=` with an absolute path, which is why I'm proposing this:
`CMAKE ?= $(shell which cmake)`

I'm using GNU Make v4.3, cmake version 3.18.4, on Debian 11.

### DIFF
--- a/build/Makefile
+++ b/build/Makefile
@@ -38,9 +38,10 @@ endif
 DESTDIR ?=
 
 
-.PHONY: faust install uninstall osc http package wasmglue
+.PHONY: faust install uninstall osc http package wasmglue cmake
 
 MAKE ?= make
+CMAKE ?= $(shell which cmake)
 
 WORKLET ?= no
 WORKLETOPT = "-DWORKLET=off"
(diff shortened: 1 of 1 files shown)


sletz:
@dfober any comment on that?

domfober:
I have doubts about the explanation for "permission denied".  I'm fine with the PR. Did it solves your problem ? (I guess it did)

MartinKlang:
I don't entirely understand it either tbh but yes, it solves my problem.

PR #741, opened 2022-06-02
Links the right sndfile library depending on the architecture on whic…

Xav83:
Links the right sndfile library depending on the architecture on which we want to build for.
Discussed on Slack with @sletz

## Bug description

When building from a M1 computer, the script was selecting the arm version of the sndfile library to be use in order to build the versions arm **and** intel for the universal binary. But linking the arm version of the sndfile library to the intel version in the universal binary triggered some erros.


sletz:
Thanks!

PR #742, opened 2022-06-03
Adds a new option in the faust2max6 to be able to generate a universa…

Xav83:
Adds a new option in the faust2max6 to be able to generate a universal external for arm and intel architecture, without being restrained by the machine on which the script is runned.

Discussed with @sletz on Slack

## Context

Before this modificotion, the script `faust2max6` was retraining the generation of the external to be for intel processor only, when the script was run on a intel computer. This restriction is in place for retrocompatibility reasons. Adding a new option would allow to keep the retrocampatibility intact while adding the possibily to users to not be restrained on intel computers.

### DIFF
--- a/tools/faust2appls/faust2max6
+++ b/tools/faust2appls/faust2max6
@@ -22,6 +22,7 @@ POLY=""
 NOPATCH="0"
 NVOICES=-1
 EFFECT=""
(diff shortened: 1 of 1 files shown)


sletz:
Thanks !

PR #745, opened 2022-06-20
Support vcvrack2

o1lo01ol1o:
This updates the vcvrack generator tool to properly set the major version of `plugin.json` to 2.


sletz:
Thanks. Finally done in a more general with a new `-version `option, see https://github.com/grame-cncm/faust/commit/3e86064509a0fb0745246b30da8f48fca0c6f79e and https://github.com/grame-cncm/faust/tree/master-dev/architecture/vcvrack.

PR #775, opened 2022-08-02
New example - gameaudio/fire

mryndzionek:
Just a small example. Result of two evenings of experimentation. I think the result is pretty good taking into account the size and complexity of the file.

### DIFF
--- a/examples/gameaudio/fire.dsp
+++ b/examples/gameaudio/fire.dsp
@@ -1,0 +1,46 @@
+import("stdfaust.lib");
+
+//----------------------------`fire`---------------------------
+// fire(is_wet) : produces burning fire sound
+//
+// #### Usage
+//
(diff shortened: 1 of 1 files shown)


sletz:
Nice ! Can you possibly add documentation the ways other examples (or at least some...) do ? (https://github.com/grame-cncm/faust/blob/master-dev/examples/gameaudio/bubble.dsp#L8)

mryndzionek:
My example is not that 'parametrized', so I'm not sure what to document.

mryndzionek:
Okay, I've refactored slightly, added on parameter and some docs.

mryndzionek:
Okay, I've refactored slightly, added on parameter and some docs.

sletz:
Thanks!

PR #776, opened 2022-08-03
New example - gameaudio/cricket

mryndzionek:
The sound is based on an example from the book "Designing Sound" by Andy Farnell. Better triggering system is added.


sletz:
Thanks!

PR #778, opened 2022-08-06
export getUserData with LIBFAUST_API

DBraun:
This change is helping with the Python API.

### DIFF
--- a/architecture/faust/dsp/libfaust-signal-c.h
+++ b/architecture/faust/dsp/libfaust-signal-c.h
@@ -86,6 +86,15 @@ extern "C"
      * @return a string representation of a signal.
      */
     LIBFAUST_API const char* Ctree2str(Signal s);
+
+    /**
+     * Return the xtended type of a signal.
+     *
+     * @param s - the signal whose xtended type to return
+     *
+     * @return a pointer to xtended type if it exists, otherwise nullptr.
+     */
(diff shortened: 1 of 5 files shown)


sletz:
Thanks but several remark here
- `getUserData` has to be added to the public `libfaust-signal.h` and `libfaust-signal-c.h` API headers
- `getUserData` currently returns a `void*` that is actually corresponding to a` xtended*` type. You will have to access this ` xtended*` data type right? So I guess we'll have to export some additional fonctions to do that?

DBraun:
Thanks. My code was working because I was including "tree.hh". Maybe I should stop including that?

I will probably want to add code related to xtended type at some point, but for now this is sufficient. When I parse this code
`process = _, 0.5 : min;`
it ends up going to `getUserData` like this https://github.com/grame-cncm/faust/blob/71d87c0ae9d9ab81001ae4ba329a4102cd5d7870/compiler/transform/sigIdentity.cpp#L60-L65
The branches are 0.5 and _, while the `sig->node()` is `min`. In Python I'm able to call `branches()`, so it's straightforward to recreate what this sigIdentity transform is doing.

sletz:
> Thanks. My code was working because I was including "tree.hh". Maybe I should stop including that?

Yes., The point is to add needed things in **public** headers only, and consider all other as internal.

PR #796, opened 2022-08-30
Add support for Trill Craft Sensors

PascalFaivre:
Individual pin :
[TRILL:CRAFT_n PIN no_pin] (return the pressure of the pin)
array of pins : 
[TRILL:CRAFT_n UP lo_pin-hi_pin] (return the index of the higher pin active in the array of pin defined or -1 if no pin active)
[TRILL:CRAFT_n DOWN lo_pin-hi_pin] (return the index of the lower pin active in the array of pin defined or -1 if no pin active)

It is fonctionnal, I leaved the request in draft because the CRAFT sensor must have offen sensibility settings defined and I would like to discuss the right place in the code for these... currently I have placed them in the sensor initialization loop, but it's probably not a good place.


sletz:
I see a lot of intermediate commits here. Can you possibly prepare a fresh PR when only the relevant changes? Thanks.

PR #798, opened 2022-09-01
Add support for Trill Craft sensors in faust2bela

PascalFaivre:
Individual pin :
[TRILL:CRAFT_n PIN no_pin] (return the pressure of the pin)
array of pins :
[TRILL:CRAFT_n UP lo_pin-hi_pin] (return the index of the higher pin active in the array of pin defined or -1 if no pin active)
[TRILL:CRAFT_n DOWN lo_pin-hi_pin] (return the index of the lower pin active in the array of pin defined or -1 if no pin active)

It is fonctional, I leaved the request in draft because the CRAFT sensor must have offen sensibility settings defined and I would like to discuss the right place in the code for these... currently I have placed them in the sensor initialization loop, but it's probably not a good place.


sletz:
Merge after some reformatting [here](https://github.com/grame-cncm/faust/commit/7759f77075854cef90e70042596708c1d44a1073). 

>currently I have placed them in the sensor initialization loop, but it's probably not a good place.

What would be the alternative?

PascalFaivre:
`What would be the alternative?`

I was thinking either of directives in the faust code like : `declare config "TRILL:CRAFT_n prescaler 4`.
(the advantage being that one can use the 'tobela' compilation option while refining the sensor settings)
or by adding a configuration file separate from the code. (the advantage being to be able to modify the settings of the sensors after the compilations)

sletz:
No sure to understand: can we possibly discuss that on [Faust Slack](https://faust.grame.fr/community/help/#faust-slack-channel)  ?

PR #799, opened 2022-09-14
JAX Backend

DBraun:
I've been using this backend in several toy experiments with JAX. Before I polish and publish those experiments, I want to start the discussion of a PR for just the backend. I don't think anything here is broken, but I would welcome suggestions especially on `instructions_compiler.cpp`. The goal there was to build text like `return state, jnp.stack([_result0, _result1, _result2])`. Getting the PR merged soon would also help me get JAX integrated into DawDreamer.

Tagging @khiner in case you're interested too!

PR #800, opened 2022-09-14
JAX Backend

DBraun:
Rebased version of https://github.com/grame-cncm/faust/pull/799


compiler/generator/instructions_compiler_jax.cpp
```diff
@@ -0,0 +1,320 @@
...
+            return generateCacheCode(sig, InstBuilder::genLoadArrayStructVar(vname, ridx2));
+        }
+    }
+}
+
+
+/**
+ * Generate code for a projection of a group of mutually recursive definitions
+ */
+ValueInst* InstructionsCompilerJAX::generateRecProj(Tree sig, Tree r, int i)
```
sletz, line 87: Is this method JAX specific ? Seems not, then it should not be here right ?

compiler/generator/instructions_compiler_jax.cpp
```diff
@@ -0,0 +1,320 @@
...
+    } else {
+        res = InstBuilder::genNullValueInst();  // Result not used
+    }
+    return res;
+}
+
+/**
+ * Generate code for a group of mutually recursive definitions
+ */
+ValueInst* InstructionsCompilerJAX::generateRec(Tree sig, Tree var, Tree le, int index)
```
sletz, line 106: Is this method JAX specific ? Seems not, then it should not be here right ?

compiler/generator/instructions_compiler_jax.cpp
DBraun: Correct, removed in latest.

compiler/generator/instructions_compiler_jax.cpp
DBraun: Correct, removed in latest.

DBraun:
There are some related things I'll have to do.

compiler/generator/jax/jax_code_container.cpp
```diff
@@ -0,0 +1,411 @@
...
+        for (const auto& it : fGlobalDeclarationInstructions->fCode) {
+            if (dynamic_cast<DeclareVarInst*>(it)) {
+                it->accept(&initializer);
+            }
+        }
+        tab(n + 2, *fOut);
+        *fOut << "# inline subcontainers:";
+        tab(n + 2, *fOut);
+        gGlobal->gJAXVisitor->Tab(n + 2);
+        inlineSubcontainersFunCalls(fStaticInitInstructions)->accept(gGlobal->gJAXVisitor);
```
DBraun, line 182: I'm testing `import("stdfaust.lib"); process = os.osc(440.);`

At this point I get code like
```python
# inline subcontainers:
l0_re0 = 0
while (l0 < 2):
	state["iVec0"]  = state["iVec0"].at[l0].set(0) 
	l0 = (l0 + 1) 
```
but `l0_re0` should be `l0` and I haven't figured out how to fix it.

compiler/generator/instructions_compiler_jax.cpp
```diff
@@ -0,0 +1,194 @@
...
+            pushComputeDSPMethod(InstBuilder::genDecStackVar(vname, InstBuilder::genBasicTyped(ctype), exp));
+        } else {
+            pushPreComputeDSPMethod(InstBuilder::genDecStackVar(vname, InstBuilder::genBasicTyped(ctype),
+                                                                InstBuilder::genTypedZero(ctype)));
+            pushComputeDSPMethod(InstBuilder::genControlInst(ccs, InstBuilder::genStoreStackVar(vname, exp)));
+        }
+
+    } else if (mxd < gGlobal->gMaxCopyDelay) {
+        // Generates table init
+        pushClearMethod(generateInitArray(vname, ctype, mxd + 1));
```
DBraun, line 63: generateInitArray is what's generating `l0_re0` instead of `l0`

DBraun:
I don't think there bugs now, but it's missing a feature I'd like. The feature would be to avoid using static tables when possible. This would speed up JAX compilation time immensely and improve the gradient differentiation with JAX.

This code
```faust
import("stdfaust.lib");
process = os.osc(440.);
```
ends up creating this Python (some parts excluded here):
```python
def classInit(self, state):
	# global declarations:
	state["fRec1"] = jnp.zeros((2,), dtype=jnp.float32)
	state["iVec0"] = jnp.zeros((2,), dtype=jnp.int32)
	state["iRec0"] = jnp.zeros((2,), dtype=jnp.int32)
	state["ftbl0MyDSPSIG0"] = jnp.zeros((65536,), dtype=jnp.float32)
	# inline subcontainers:	
	for i1 in range(0, 65535):
		state["iVec0"] = state["iVec0"].at[0].set(1) 
		state["iRec0"] = state["iRec0"].at[0].set(((state["iVec0"][1] + state["iRec0"][1]) % 65536)) 
		state["ftbl0MyDSPSIG0"]  = state["ftbl0MyDSPSIG0"].at[i1].set(jnp.sin((9.58738e-05 * (state["iRec0"][0])))) 
		state["iVec0"] = jnp.roll(state["iVec0"], 1) 
		state["iRec0"] = jnp.roll(state["iRec0"], 1) 

@staticmethod
def tick(state: dict, inputs: jnp.array):
		
	state["fRec1"] = state["fRec1"].at[0].set((state["fConst0"] + (state["fRec1"][1] - jnp.floor((state["fConst0"] + state["fRec1"][1]))))) 
	_result0 = state["ftbl0MyDSPSIG0"][jnp.int32((65536.0 * state["fRec1"][0]))] 
	state["fRec1"] = jnp.roll(state["fRec1"], 1) 
	return state, jnp.stack([_result0]) 
```

It turns out that the `for i1 in range(0, 65535)` takes over a minute to compile because those operations are slow for JAX. It would be better to just compute jnp.sin on the fly in the tick function.

DBraun:
What I said above can be put on hold because one can just code an oscillator that doesn't use rdtable. I copied from the Box Tutorial:
```faust
import("stdfaust.lib");
process = osc(f1), osc(f2)
with {
    decimalpart(x) = x-int(x);
    phasor(f) = f/ma.SR : (+ : decimalpart) ~ _;
    osc(f) = sin(2 * ma.PI * phasor(f));
    f1 = vslider("Freq1", 300, 100, 2000, 0.01);
    f2 = vslider("Freq2", 500, 100, 2000, 0.01);
};
```
And that compiles quickly. But if something else in faustlibraries calls os.osc, then I would want a way to use a non-rdtable version.

DBraun:
I have some tests here: https://gist.github.com/DBraun/556024cb51ecc7416cc7c9509c7a5259

`python -m pytest -v test_jax_backend.py`

They all pass, but I would need to add a python continuous integration thing to the faust repo.

DBraun:
Should I add a Python installer and test to Ubuntu (https://github.com/grame-cncm/faust/blob/master-dev/.github/workflows/ubuntu.yml)? It should be easy to install JAX on Ubuntu.

sletz:
What are the [jax-tests](https://github.com/DBraun/faust/tree/jax-backend-2/tests/jax-tests) actually testing ?

DBraun:
Every function that starts with “test” inside test_jax_backend.py would be run. Each of those functions has some Faust code that gets converted to JAX, built with the JAX minimal.py architecture, and run. So it tests that the JAX backend works and that the JAX code doesn’t throw errors.

sletz:
Backends are tested in the tests/impulse-teste infrastructures (so not at each commit). How complex would it be to move the JACK backend tests here?

DBraun:
It would be easy to change the relative paths in test_jax_backend.py. So would you suggest moving the jax-tests to tests/interp-tests/jax-texts?

DBraun:
The current progress:

Two files don't compile:

`osc_enable.dsp` doesn't work because it results in code with `if` statements inside its one-sample block.

```python
if (iTemp1 != 0):
    fTemp3 = (fTemp0 * fTemp2) 
```

Any tips for generating code that doesn't do this? I can imagine getting similar results if the visitor went through `visit(Select2Inst* inst)`

The other file that doesn't compile is `bs.dsp`. I could eventually get this working, but it's kind of low priority in my opinion.

If I set the `filesCompare` precision to 2, ignoring `bs.dsp` and `osc_enable.dsp`, all the tests pass. There are some examples which don't pass low precision differences, such as `carre_volterra.dsp`:
```
/Applications/Xcode.app/Contents/Developer/usr/bin/make -f Make.jax outdir=jax/double   FAUSTOPTIONS="-I dsp -double"
python3  ir/jax/double/jax_carre_volterra.py > ir/jax/double/carre_volterra.ir
./filesCompare ir/jax/double/carre_volterra.ir reference/carre_volterra.ir 1e-4       
Line : 2632 output : 0 sample1 : 0.090985 different from sample2 : 0.091085 delta : 0.0001
Line : 2633 output : 0 sample1 : 0.056059 different from sample2 : 0.056159 delta : 0.0001
Line : 2681 output : 0 sample1 : -0.125745 different from sample2 : -0.125846 delta : 0.000101
Line : 2682 output : 0 sample1 : -0.091029 different from sample2 : -0.091131 delta : 0.000102
Line : 2683 output : 0 sample1 : -0.056184 different from sample2 : -0.056285 delta : 0.000101
Line : 2684 output : 0 sample1 : -0.021322 different from sample2 : -0.021424 delta : 0.000102
Line : 2685 output : 0 sample1 : 0.013448 different from sample2 : 0.013347 delta : 0.000101
Line : 2686 output : 0 sample1 : 0.048029 different from sample2 : 0.047928 delta : 0.000101
Line : 2687 output : 0 sample1 : 0.082328 different from sample2 : 0.082228 delta : 0.0001
Line : 2731 output : 0 sample1 : 0.125626 different from sample2 : 0.125727 delta : 0.000101
Line : 2732 output : 0 sample1 : 0.090992 different from sample2 : 0.091094 delta : 0.000102
Line : 2733 output : 0 sample1 : 0.056222 different from sample2 : 0.056325 delta : 0.000103
```

Should I be testing `-double`? What are your suggestions for debugging small numerical differences?

DBraun:
Precision tests all pass! I only have to ignore bs.dsp and osc_enable.dsp in `Make.jax`.

sletz:
String to look at the code; It would be simpler for me to comment the code after 1) rebase on master-dev 2) and prepare a PR in --squash mode (that is a unique single big commit). Thanks.

sletz:
Thanks for the rebase. I will do general comments on the PR here, and specific ones directly in the code.
General comments looking at the generated code:
- I understand the all "init" like functions (`init/instanceInit/classInit` etc...) are grouped in a single `classInit`. The usual somatic for `classInit` in the C++ generated code is to share init code that is somewhat commit to all "instances". Here I don't think you really have this "static init" versus "instance init" anymore. So I would prefer having `classInit` simply become `init`
- all those` load_soundfile/add_soundfile/add_nentry/add_button/add_slider ` are just generic code right ? Then why having the backend generate them? Wouldn't be better to have them part of an architecture file ? (which is the standard way todo...)
-  when compiling freeverb.dsp , I see: `state["IOTA0"] = 0 ` generated in `build_interface` . This seems like an initialisation right? So better move that in `init`
- still an indentation issue in the generated code (probably an non match "tab" at the end of the file)

architecture/jax/minimal.py
```diff
@@ -0,0 +1,85 @@
...
+# FAUST architecture file. Redistribution and use in source and binary
+# forms, with or without modification, in part or in full are permitted.
+# In particular you can create a derived work of this FAUST architecture
+# and distribute that work under terms of your choice.
+
+# This sample code is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+# ************************************************************************
+
```
sletz, line 16: Could this file contain `load_soundfile/add_soundfile/add_nentry/add_button/add_slider` functions?

compiler/extended/powprim.hh
```diff
@@ -146,16 +146,18 @@ class PowPrim : public xtended {
         if (isIntPowArg(types[1], *it, pow_arg)
             && (types[1]->variability() == kKonst)
             && (types[1]->computability() == kComp)
-            && (gGlobal->gNeedManualPow)) {
+            && (gGlobal->gNeedManualPow)
+            && (gGlobal->gOutputLang != "jax")) {
```
sletz, line 150: What that?

compiler/generator/instructions_compiler.cpp
```diff
@@ -546,6 +553,10 @@ void InstructionsCompiler::compileMultiSignal(Tree L)
         }
     }
 
+    // these two vars are only used for jax
```
sletz, line 556: I guess this should be wrapped in a `if (gGlobal->gOutputLang == "jax") {...}` right ?

compiler/generator/instructions_compiler_jax.cpp
```diff
@@ -0,0 +1,178 @@
...
+ along with this program; if not, write to the Free Software
+ Foundation, Inc., 675 Mass Ave, Cambridge, MA 02139, USA.
+ ************************************************************************
+ ************************************************************************/
+
+#include "instructions_compiler_jax.hh"
+#include "ensure.hh"
+#include "ppsig.hh"
+#include "sigtyperules.hh"
+
```
sletz, line 26: Be sure to check that all methods here are necessary, that is are **really** JAX backend specific.

compiler/generator/jax/jax_code_container.cpp
```diff
@@ -0,0 +1,486 @@
...
+        }
+        tab(n + 2, *fOut);
+        *fOut << "json_str = \"\"\"" << flattenJSONforPython(json) << "\"\"\"";
+        tab(n + 2, *fOut);
+        *fOut << "return json.loads(json_str)";
+        tab(n + 1, *fOut);
+    }
+
+    tab(n + 1, *fOut);
+    *fOut << "def load_soundfile(self, filepath):";
```
sletz, line 252: General remark on the necessity to have the backend generate `load_soundfile/add_soundfile/add_nentry/add_button/add_slider`

compiler/generator/jax/jax_instructions.hh
```diff
@@ -0,0 +1,803 @@
...
+     Global functions names table as a static variable in the visitor
+     so that each function prototype is generated as most once in the module.
+     */
+    static map<string, bool> gFunctionSymbolTable;
+
+    // Polymorphic math functions
+    map<string, string> gPolyMathLibTable;
+    
+    bool fMutateFun;
+    bool is_storing_lhs = false;
```
sletz, line 129: Can you follow the naming/formatting code convention we have for class field ? So  `is_storing_lhs` would become `fIsStoringLhs` and same for other ones.

compiler/global.hh
```diff
@@ -566,6 +571,7 @@ struct global {
     int    gVHDLFloatMSB;
     int    gVHDLFloatLSB;
     bool   gElementarySwitch;
+    bool   gJAXSwitch;
```
sletz, line 574: This seem not necessary anymore ?

compiler/libcode.cpp
```diff
@@ -401,6 +410,10 @@ static bool processCmdline(int argc, const char* argv[])
             gGlobal->gElementarySwitch = true;
             i += 1;
 
+        } else if (isCmd(argv[i], "-jax", "--jax")) {
```
sletz, line 413: Not necessary anymore ?

DBraun (commented):
Sorry I meant to "submit" the review earlier.

compiler/extended/powprim.hh
sletz: Setting `gGlobal->gNeedManualPow` to true/false is something each backend can decide in lib code.cpp, see for instance: https://github.com/grame-cncm/faust/blob/master-dev/compiler/libcode.cpp#L1435

sletz:
Looking at the generated code: call` state = self.build_interface(state, x, T)` seems like somewhat strange and does not really follow "blue/red/green" model of architecture https://faustdoc.grame.fr/manual/architectures/#architecture-files
Why not removing `build_interface` (and have `initialize` init the DSP state but not more) and have:
```
def __call__(self, x, T: int) -> jnp.array:
   state = self.initialize(self.sample_rate, x, T)
   state = self.build_interface(self.sample_rate, x, T)
   return jnp.transpose(jax.lax.scan(self.tick, state, jnp.transpose(x, axes=(1, 0)))[1], axes=(1,0))
```
Ot is is because `self.build_interface`is considered as initialising the DSP state in this JAX backend ?

sletz:
Could you possibly document a bit `jax_code_container.cp` with the JAX backend particularities as in https://github.com/grame-cncm/faust/blob/master-dev/compiler/generator/julia/julia_code_container.cpp#L32 ?

sletz:
It seems  `bool fMutateFun;` in JAXInstVisitor is not used anymore ?

sletz:
In JAXInstVisitor, please use the opcode name instead if `7` in the following line:
`if (inst->fOpcode > 7 && !fIsDoingWhile) {`

DBraun:
The code is now more like

```python
def __call__(self, x, T: int) -> jnp.array:
   state = self.initialize(self.sample_rate, x, T)
   state = self.build_interface(state, x, T)
   state = jax.tree_map(jnp.array, state)
   return jnp.transpose(jax.lax.scan(self.tick, state, jnp.transpose(x, axes=(1, 0)))[1], axes=(1,0))
```
The `state` needed to be handed off to `build_interface`.

I would like to get the `add_soundfile` `add_slider` etc functions into an architecture file, but I didn't figure out how to default to a specific architecture file if none is specified.

I removed `bool fMutateFun;`, added some comments in `jax_code_container.cpp` and improved `if (inst->fOpcode > 7 && !fIsDoingWhile) {`

Can you tell me where `FAUSTFLOAT(foo)` is called? I had to define a dummy function

```python
def FAUSTFLOAT(x):
    return x
```

But it would be better if `FAUSTFLOAT(foo)` was just `foo` everywhere. I didn't figure it out with breakpoints much earlier.

sletz:
"but I didn't figure out how to default to a specific architecture file if none is specified." => well even in C++, the compile actually the compiled does not generated a self-contained class.... It always assumes `dsp/UI/meta` etc... are defined in and architecture files. This is way I think we could just do the same same with JAX and assume (also..) that te appropriate architecture will be used.

DBraun:
Good point. I'll try that tomorrow.

sletz:
`Can you tell me where FAUSTFLOAT(foo) is generated in C++?`  => I"ll check that

sletz:
`FAUSTFLOAT`  machinery makes sense in C/C++. With JAX I think you should do as with the LLVM backend where we actually consider that `FAUSTFLOAT` is the same type as internal REAL type (that is either `float` or `double` depending if `-single` (= default) or `-double` is used as an option). See https://github.com/grame-cncm/faust/blob/master-dev/compiler/global.hh#L166 and https://github.com/grame-cncm/faust/blob/master-dev/compiler/libcode.cpp#L1396.

DBraun:
I moved most of the code into the architecture files now. I think setting `gGlobal->gFAUSTFLOAT2Internal  = true;` helped avoid FAUSTFLOAT being called.

DBraun (commented):
Tried moving code into architecture files.

compiler/generator/instructions_compiler_jax.cpp
DBraun: Yes I’ve checked again.

compiler/generator/jax/jax_code_container.cpp
DBraun: Should be resolved by recent commits.

sletz:
Great work ! Merged in --squash mode after some final cleanup and reformatting in https://github.com/grame-cncm/faust/commit/44d66aac61b05cb172e101a2b4051e2aa0ea248f

Still 2 thinks to check:
- Is this `FAUSTFLOAT` related block still necessary https://github.com/grame-cncm/faust/commit/44d66aac61b05cb172e101a2b4051e2aa0ea248f#diff-5ac628298a99ca6f354d9f3f5a64424743f37be2451f2ab84e6645418f1c02e4R149-R165 ?
- the `JAXCodeContainer::inlineSubcontainersFunCalls ` is almost the same the the base class `CodeContainer::inlineSubcontainersFunCalls` (with only the renaming block changing AFAICS) https://github.com/grame-cncm/faust/commit/44d66aac61b05cb172e101a2b4051e2aa0ea248f#diff-5ac628298a99ca6f354d9f3f5a64424743f37be2451f2ab84e6645418f1c02e4R296. Is renaming an issue here? If not then I guess the base class `CodeContainer::inlineSubcontainersFunCalls` should be used.

sletz:
Now everything is OK, thanks !

PR #803, opened 2022-09-18
Rust: Allocate arrays on the heap

plule:
Even though the code wraps the allocation of the DSP in a `Box<>`, the defined behaviour of rust is that the memory is first allocated in the stack, then moved in the heap. In Release build, this is often optimized out, hiding this issue, but this makes Faust generated rust code very prone to stack overflow in Debug. This is generally considered an issue in Rust, see: https://github.com/rust-lang/rust/issues/53827

It looks like a rust unstable feature (`box` syntax) would allow to allocate directly on the heap with relying on optimization, but this has not been stabilized. In the meantime, using `Vec` seems to be the general advise for this problem, so this PR replaces the arrays with vectors.

### DIFF
--- a/compiler/generator/rust/rust_instructions.hh
+++ b/compiler/generator/rust/rust_instructions.hh
@@ -60,9 +60,9 @@ struct RustInitFieldsVisitor : public DispatchVisitor {
 
         if (array_type) {
             if (isIntPtrType(type)) {
-                *fOut << "[0;" << array_type->fSize << "]";
+                *fOut << "vec![0;" << array_type->fSize << "]";
             } else if (isRealPtrType(type)) {
-                *fOut << "[0.0;" << array_type->fSize << "]";
+                *fOut << "vec![0.0;" << array_type->fSize << "]";
             }
         } else {
             if (isIntType(type)) {
--- a/compiler/generator/type_manager.hh
+++ b/compiler/generator/type_manager.hh
@@ -232,9 +232,13 @@ class RustStringTypeManager : public StringTypeManager {
             string ty_str = named_typed->fName + generateType(named_typed->fType);
             return name + ((ty_str != "") ? (": " + ty_str) : "");
         } else if (array_typed) {
+            // Allocate arrays on the heap with std::vec::Vec
+            // Until the rust feature box_syntax is stabilized, this is necessary
+            // to avoid data to be allocated first on the stack, then moved to the heap.
+            // See: https://github.com/rust-lang/rust/issues/53827
             return (array_typed->fSize == 0)
                        ? name + ": " + fPtrRef + generateType(array_typed->fType)
-                       : name + ": [" + generateType(array_typed->fType) + ";" + std::to_string(array_typed->fSize) + "]";
+                       : name + ": std::vec::Vec<" + generateType(array_typed->fType) + ">";
         } else {
             faustassert(false);
             return "";


bluenote10:
To simplify reviewing, could you perhaps show some before/after of code generation? Would this for instance mean that in the following example dsp all arrays would be replaced by vectors?

```rust
pub struct MyDsp {
    fHslider0: f32,
    fHslider1: f32,
    fVec0: [f32; 3],
    fHslider2: f32,
    fSampleRate: i32,
    fConst0: f32,
    fConst1: f32,
    fConst2: f32,
    fHslider3: f32,
    fRec25: [f32; 2],
    fRec29: [f32; 2],
    fRec31: [f32; 2],
    fVec1: [f32; 3],
    IOTA: i32,
    fRec32: [f32; 8192],
    fVec2: [f32; 2],
    fHslider4: f32,
    iRec34: [i32; 2],
    fConst3: f32,
    fRec33: [f32; 3],
    fVec3: [f32; 2],
    iRec35: [i32; 2],
    fConst4: f32,
    fVec4: [f32; 3],
    fRec30: [f32; 8192],
    fRec21: [f32; 2],
    fRec17: [f32; 2],
    fRec13: [f32; 8192],
    fRec15: [f32; 2],
    fRec11: [f32; 2],
    fRec6: [f32; 2],
    fRec2: [f32; 8192],
    fRec0: [f32; 2],
}
```

I vaguely remembering considering this as well, but decided against because it, probably because:
- It would enforce heap usage => showstopper for embedded use case?
- Heap fragmentation due to many tiny vecs.
- Hope that box syntax gets stabilized soon (https://github.com/rust-lang/rust/issues/49733).
- Observation that release builds already do what the box syntax will deliver officially.
- For debug builds, temporarily increasing the stack can serve as an ugly work-around.

Of course, relying on optimization isn't great, but enforcing heap usage isn't great either :confused: 

I could imagine that putting all the `fRec` on the heap will have a quite negative impact on cache efficiency. It would be interesting to re-run the benchmarks at https://github.com/bluenote10/RustFaustExperiments/tree/master/Benchmarks to see how bad it would be. Perhaps we should only move buffers above a certain size to the heap? (Perhaps that what the PR already does?)

plule:
> To simplify reviewing, could you perhaps show some before/after of code generation? Would this for instance mean that in the following example dsp all arrays would be replaced by vectors?

That's right, all the arrays are replaced with vectors. There is nothing smart with small vecs in the PR, so this includes the arrays of 2 and 3.

> It would enforce heap usage => showstopper for embedded use case?

Yep :( I actually did not consider that at all, my reasoning was "well, since the DSP is put on the heap, let's ensure it works in all cases". Though I assume people working in the embedded don't use the architecture files `Box`ing the DSP. This point alone convinced me that this approach is not the right one...

I will try to confirm that the new syntax is indeed fixing the issue.

plule:
Well well, from my attempts the new box syntax does not solve the problem, it only prevent stack allocation when directly boxing the arrays, not boxing a structure with big arrays... I think that the only solution keeping all the advantages involve unsafe allocation. I'll open an issue to discuss other solutions, this one is not the right approach.

smoothdeveloper:
@plule, I faced issue with allocating large arrays in Rust, the solution has been to resort on the `std::Vec` datastructure which you can initialise to a specified capacity (https://doc.rust-lang.org/std/vec/struct.Vec.html#method.with_capacity) and will handle heap allocation.

plule:
Yes @smoothdeveloper , I did not post the generated code, but this is what this PR did actually... Unfortunately there are the issues pointed above that makes this unsuitable for this particular case.

PR #806, opened 2022-09-24
Daisy Pod support & for some bugfix.

tomoyanonymous:
This PR fixes some compilation errors of faust2daisy on my environment and adds basic support for Daisy Pod based on a current implementation for a Patch.
The version of my libdaisy is a current master branch.

-  `static_cast` added on this commit needed to be`const_cast` in the latest libdaisy. https://github.com/grame-cncm/faust/commit/a0e58b5a4aecc9c1d3d6382140aa82a14ba21f98#diff-db20fb08bb6609ca9435273eefd9ec4f826a0116d40cd83049c507415dc417abR100
- `hw.adc.Start();` for seed needs to be `hw.StartAdc();` for Patch and Pod.
- The pin assignments of Knobs are different for each platform. 
  - Pod: `15 21`
  - Patch: `15 16 21 18`
  - Petal :`16 19 17 20 18 21`
Pod users can use only `knob:1` and `knob:3` to use the common pin mappings to Patch.
It would be better to define correct pin assignments in the preprocessor, but I added a description of knob mappings in the help of `faust2daisy` for now.

Maybe I can add Encoder support and Petal platform support in the future.

### DIFF
--- a/architecture/daisy/ex_faust.cpp
+++ b/architecture/daisy/ex_faust.cpp
@@ -36,6 +36,8 @@
 
 #ifdef PATCH
 #include "daisy_patch.h"
+#elif defined POD
+#include "daisy_pod.h"
 #else
 #include "daisy_seed.h"
 #endif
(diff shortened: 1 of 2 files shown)


sletz:
Thanks !

PR #807, opened 2022-09-29
fix: ladspa architecture file to set descriptor info from metadata

SpotlightKid:
* Set LADSPA descriptor fields `Name`, `Maker`, `Copyright` from FAUST global metadata `name`, `author`, `copyright`/`license` fields, if present.
* Set `Maker` to empty string, not `"undefined"`, if not present.

Note: I'm no C/C++ expert, but I took care copy the descriptor info from the global FAUST metadata in a safe way. Please review the changes with this in mind.

### DIFF
--- a/architecture/ladspa.cpp
+++ b/architecture/ladspa.cpp
@@ -80,6 +80,7 @@
 //--------------------------------useful constants--------------------------------------
 
 #define MAXPORT 1024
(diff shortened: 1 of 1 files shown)


SpotlightKid:
Sorry, I messed up the changes in this PR.

I created #808 to fix this.

PR #809, opened 2022-09-29
Enable Interpreter Backend on Windows

romsom:
This PR enables the interpreter backend on Windows platforms. An appropriate code path has existed before, but compilation was disabled.
I noticed the missing function symbols during testing of my integration project of Faust into [HISE](https://github.com/christophhart/HISE/) which I'm working on as a Google Summer of Code contributor this year.

I discussed the necessary changes with @sletz and we found the ad-hoc solutions presented in this PR. Better solutions will probably have to be found for some of then, especially the disabling of several warnings, which had used GCC built-ins, which are unavailable on MSVC. Still I wanted to send the changes in as a template.


sletz:
Reworked and merged in --squash mode here: https://github.com/grame-cncm/faust/commits/master-dev Can you test? Thanks.

romsom:
Works here, thanks!

PR #811, opened 2022-10-05
add SLIDER mode for trill craft sensor

PascalFaivre:
consist of simulate a slider with 2 pins of a craft sensor (one for increase and on for decrease the value)


architecture/bela.cpp
```diff
@@ -495,6 +503,9 @@ class BelaWidget
         ,fLabel(w.fLabel)
         ,fMin(w.fMin)
         ,fRange(w.fRange)
+        ,fbuffered(w.fbuffered)
+        ,fbuffer(0)
+        ,old_state(0)
```
sletz: Coule you follow the coding convention here:` old_state -> fOldState`

architecture/bela.cpp
```diff
@@ -478,6 +480,9 @@ class BelaWidget
         const char* fLabel;  // label
         FAUSTFLOAT fMin;    // minimal value
         FAUSTFLOAT fRange;  // value range (max-min)
+        bool fbuffered;
+        double fbuffer;
```
sletz: Is is `double` or `FAUSTFLOAT` here?

sletz:
Merged in --squash mode after a bit or reformatting (class fields renaming...etc..) in https://github.com/grame-cncm/faust/commit/2656d964ba2525690dfafa287ec23285b053cf79. Can you check everything still works?

PR #819, opened 2022-10-16
Fix accidental clashing intermediate JAX variables

DBraun:
There's a problem with the current code at this line:
https://github.com/grame-cncm/faust/blob/3af02793c9b6c2244f6e9d38082819b511936d80/architecture/jax/minimal.py#L111

The label isn't unique so multiple things try to write to the same intermediate variable. I'm talking about the intermediate variables `mod_vars` [here](https://github.com/grame-cncm/faust/blob/3af02793c9b6c2244f6e9d38082819b511936d80/architecture/jax/minimal.py#L157).

Consider this Faust code:
```faust
import("stdfaust.lib");

// Define the number of sections
N = 3;

// Define the parameters for each section.
gains = 0, 0, 0;
freqs = 500, 5000, 10000;
bandwidths = 100, 5000, 5000;

section(i) = hgroup("Section %i",
    fi.peak_eq(
        vslider("Gain", gains : ba.selector(i, N), -80., 12., .001),
        vslider("Freq",  freqs : ba.selector(i, N), 20., 20000., .001),
        vslider("Bandwidth", bandwidths : ba.selector(i, N), 20., 5000., .001)
    )
);

// Define a final volume slider.
volume = _*(vslider("Volume[unit:dB]", 0., -12., 12., .001) : ba.db2linear);

// Sequence all the sections.
process = hgroup("Equalizer", seq(i, N, section(i)) : volume);
```

Previously the intermediate variables would have been
```python
FrozenDict({
    Bandwidth: (DeviceArray(100., dtype=float32),),
    Freq: (DeviceArray(499.99973, dtype=float32),),
    Gain: (DeviceArray(-40., dtype=float32),),
    Volume: (DeviceArray(0., dtype=float32),),
})
```
This is wrong because it's undercounting the true number of parameters used.

The code has worked so far because what matters for the output audio is this line:

https://github.com/grame-cncm/faust/blob/3af02793c9b6c2244f6e9d38082819b511936d80/architecture/jax/minimal.py#L90

The names there end up being unique due to the use of `zone` in the generated name. So we need to do something similar for the intermediate values. What has worked in my test is to use `OpenboxInst` and `CloseboxInst` to keep track of the full path. Then the call to `self.sow('intermediates', label, param)` will actually have a unique label based on the accumulated OpenboxInst instructions.

So now the intermediate variables look like

```python
FrozenDict({
    Equalizer/Section 0/Bandwidth: (DeviceArray(100., dtype=float32),),
    Equalizer/Section 0/Freq: (DeviceArray(499.99973, dtype=float32),),
    Equalizer/Section 0/Gain: (DeviceArray(-40., dtype=float32),),
    Equalizer/Section 1/Bandwidth: (DeviceArray(5000., dtype=float32),),
    Equalizer/Section 1/Freq: (DeviceArray(5000., dtype=float32),),
    Equalizer/Section 1/Gain: (DeviceArray(3.0000021, dtype=float32),),
    Equalizer/Section 2/Bandwidth: (DeviceArray(5000., dtype=float32),),
    Equalizer/Section 2/Freq: (DeviceArray(10000., dtype=float32),),
    Equalizer/Section 2/Gain: (DeviceArray(-19.999998, dtype=float32),),
    Equalizer/Volume: (DeviceArray(0., dtype=float32),),
})
```

I also made the add_* UI functions not need to return anything. Now they modify the state dictionary that's passed as an argument.

### DIFF
--- a/architecture/jax/minimal.py
+++ b/architecture/jax/minimal.py
@@ -16,6 +16,7 @@
 
 import json
 import re
+from typing import List
 
 import numpy as np
 import librosa
@@ -43,7 +44,7 @@ def remainder(x, y):
 			audio = np.expand_dims(audio, 0)
 		return audio, sr
 	
-	def add_soundfile(self, state, x, label: str, url: str, key: str):
+	def add_soundfile(self, state, zone: str, ui_path: List[str], label: str, url: str, x):
 		# todo: better parsing
 		filepaths = url[2:-2].split("';'")
 		fLength, fOffset, fSR, offset = [], [], [], 0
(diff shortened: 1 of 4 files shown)


sletz:
Same label at different locations in the UI hierarchy is a problem which is indeed solved using the "path" (AKA "complete path") model, implemented [Path Builder](https://github.com/grame-cncm/faust/blob/master-dev/architecture/faust/gui/PathBuilder.h) and use in [MapUI](https://github.com/grame-cncm/faust/blob/master-dev/architecture/faust/gui/MapUI.h) for instance. (Note that we recently added the _short-name_ model, but this may not be relevant here). 

So I understand you're implementing something similar to the path model here right ?

DBraun:
Yes I hope it's equivalent to the path model.

This is the generated Python:
```python
        def build_interface(self, state, x, T: int):
                ui_path = []
                ui_path.append("Equalizer")
                ui_path.append("Section 0")
                self.add_slider(state, "fVslider8", ui_path, "Bandwidth", 1e+02, 2e+01, 5e+03, "linear")
                self.add_slider(state, "fVslider7", ui_path, "Freq", 5e+02, 2e+01, 2e+04, "linear")
                self.add_slider(state, "fVslider6", ui_path, "Gain", -4e+01, -8e+01, 12.0, "linear")
                ui_path.pop()
                ui_path.append("Section 1")
                self.add_slider(state, "fVslider5", ui_path, "Bandwidth", 5e+03, 2e+01, 5e+03, "linear")
                self.add_slider(state, "fVslider4", ui_path, "Freq", 5e+03, 2e+01, 2e+04, "linear")
                self.add_slider(state, "fVslider3", ui_path, "Gain", 3.0, -8e+01, 12.0, "linear")
                ui_path.pop()
                ui_path.append("Section 2")
                self.add_slider(state, "fVslider2", ui_path, "Bandwidth", 5e+03, 2e+01, 5e+03, "linear")
                self.add_slider(state, "fVslider1", ui_path, "Freq", 1e+04, 2e+01, 2e+04, "linear")
                self.add_slider(state, "fVslider0", ui_path, "Gain", -2e+01, -8e+01, 12.0, "linear")
                ui_path.pop()
                self.add_slider(state, "fVslider9", ui_path, "Volume", 0.0, -12.0, 12.0, "linear")
                ui_path.pop()

                return state
```

So `ui_path.append("Equalizer")` corresponds to `ui_interface->openHorizontalBox("Equalizer");` in a C++ architecture.

When the first add_slider is called, we're able to concatenate "Equalizer","Section 0", and "Bandwidth" to get "Equalizer/Section 0/Bandwidth". Hopefully that's equivalent to the path model.

Would you suggest using PathBuilder.h inside jax_instructions.hh?

sletz:
"_using PathBuilder.h inside jax_instructions.hh_", possibly yes, but your way is OK also, so merging.

PR #828, opened 2022-10-29
Rust: Allocate on the heap with default-boxed

### LINKED ISSUE #804: Rust: Stack overflow in debug with temporal effects
Follow-up to #803 , paging @bluenote10 and @smoothdeveloper who participated on it.

## Problem

As soon as a temporal effect is used (delay), a buffer big enough to store it is allocated in an array. Simplified generated code:

```rs
struct Dsp {
    pub a: [f32; 1024 * 1024],
}

impl Dsp {
    pub fn new() -> Self {
        Self {
            a: [0.0; 1024 * 1024],
        }
    }
}
```

This array does not fit in the stack, so allocating it fails with a stack overflow. In the architecture files, this DSP is allocated with `Box::new(Dsp::new())`. In debug compilation, the value is first allocated on the stack, then moved on the heap, leading to the same stack overflow. Fortunately in Release mode the allocation is optimized to go directly on the stack, so this problem is only appearing in debug. This problem is known in rust: https://github.com/rust-lang/rust/issues/53827

## Solution attempt 1

A possible fix is to allocate the arrays directly in a `Vec`, which is on the heap. Roughly:

```rs
struct Dsp {
    pub a: Vec<f32>,
}

impl Dsp {
    pub fn new() -> Self {
        Self {
            a: vec![0.0; 1024 * 1024],
        }
    }
}
```

This solution can be seen in #803 and solves the problem. Unfortunately, it also comes with its own problems, namely forcing heap allocation even when unnecessary, and fragmenting the memory allocation.

This approach could be revived by only allocating huge arrays in vectors, and keep smaller ones in arrays, but I feel like we would just be "chasing" the solution.

## Solution attempt 2

Second attempt: there is an unstable syntax dedicated to allocating on the heap: `box_syntax`. Unfortunately, it's been unstable for years, and at this point does not solve the stack overflow:

```rs
// Ok
let a = box [0.0; 1024 * 1024];
// Stack overflow
let dsp = box Dsp::new();
```

## Solution attempt 3

Third party crates such as [default-boxed](https://github.com/upsuper/default-boxed) solve this problem by allocating the memory on the heap, then assigning the initial values there. The implementation includes `unsafe` code (as any manual allocation), and solves the issue, while retaining the memory layout, without baking the heap allocation in the Dsp:

```rs
#[cfg_attr(feature = "default-boxed", derive(default_boxed::DefaultBoxed))]
struct Dsp {
    pub a: [f32; 1024 * 1024],
}

fn main() {
    let dsp = Dsp::default_boxed();
}
```

I think that this option is nice enough, and easy to implement. The dependency to default-boxed can be made optional by putting this behind an opt-in feature flag (which is what the cfg_attr does).

While this last solution ticks all the boxes, I'm unsure if it's worth the added complexity and possible confusion from the "optional dependency" mechanism. I'd be up to implement it in Faust, Frando/rust-faust and the architecture files if this solution is suitable.

Let me know what you think, maybe I've missed another possible approach.

plule:
Add the ability to allocate directly on the heap with default-boxed.

Fixes #804

At the moment, there is no formal way to allocate an arbitrary struct on the heap in rust. The current approach taken by Faust is to allocate the dsp with `Box::new(Dsp::new())` which would be expected to first allocate on stack then move on heap, but this move is optimized away in release and it's allocated directly on the heap. However, when running in debug mode, the intermediary allocation is here making any echo or reverb very likely to stack overflow.

This PR adds a way to allocate the DSP on the heap with the help of the third-party crate [default-boxed](https://docs.rs/default-boxed/latest/default_boxed/).

The change in the generated code of Faust is very conservative: It's under conditional compilation with a feature, disabled by default. When the feature is enabled, it adds a new `default_boxed()` method. This change should not break any user, including in specific environment such as embedded/no-std.

```rs
#[cfg_attr(feature = "default-boxed", derive(default_boxed::DefaultBoxed))]
#[derive(Debug,Clone)]
pub struct mydsp {
	iVec0: [i32;2],
```

The architecture files are updated to take advantage of it. The feature is enabled by default in `Cargo.toml` so users of the `faust2XXX` scripts will benefit of it. They can still opt-out with `cargo build --no-default-features`.

### DIFF
--- a/architecture/cpal.rs
+++ b/architecture/cpal.rs
@@ -106,7 +106,17 @@ pub trait UI<T> {
 fn main() -> Result<(), anyhow::Error> {
 
     // Allocation DSP on the heap
-    let mut dsp = Box::new(mydsp::new());
+    let mut dsp;
+    #[cfg(feature = "default-boxed")]
+    {
+        use default_boxed::DefaultBoxed;
+        dsp = mydsp::default_boxed();
+    }
+
+    #[cfg(not(feature = "default-boxed"))]
+    {
+        dsp = Box::new(mydsp::new());
+    }
 
     println!("get_num_inputs: {}", dsp.get_num_inputs());
     println!("get_num_outputs: {}", dsp.get_num_outputs());
--- a/architecture/jack.rs
+++ b/architecture/jack.rs
@@ -108,7 +108,17 @@ fn main() {
     let (client, _status) = j::Client::new("faust_rust", j::client_options::NO_START_SERVER).unwrap();
 
     // Allocation DSP on the heap
-    let mut dsp = Box::new(mydsp::new());
+    let mut dsp;
+    #[cfg(feature = "default-boxed")]
+    {
+        use default_boxed::DefaultBoxed;
+        dsp = mydsp::default_boxed();
+    }
+
+    #[cfg(not(feature = "default-boxed"))]
+    {
+        dsp = Box::new(mydsp::new());
+    }
 
     println!("Faust Rust code running with JACK: sample-rate = {} buffer-size = {}", client.sample_rate(), client.buffer_size());
 
--- a/architecture/minimal.rs
+++ b/architecture/minimal.rs
@@ -189,7 +189,18 @@ fn main() {
 
     println!("Faust Rust DSP");
 
-    let mut dsp = Box::new(mydsp::new());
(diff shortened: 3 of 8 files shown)


sletz:
Thanks!

PR #833, opened 2022-11-27
Improve dplug.d architecture file

abaga129:
This adds the following to the dplug.d architecture file

- Sorts the parameters based on their order metadata
- Makes use of the `unit` metadata
- Checks if `step == 1.0` when creating parameters to determine integer or float parameters.
- Allows the user to add `faustoverride` to their versions in `dub.json` to override the plugin entrypoints and use the output code in an existing project.

### DIFF
--- a/architecture/dplug.d
+++ b/architecture/dplug.d
@@ -36,6 +36,7 @@
 
 import dplug.core.vec;
 import dplug.client;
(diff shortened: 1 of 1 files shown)


sletz:
Thanks ! Merged after a bit of tab/space and reformatting.

PR #837, opened 2022-12-02
Support LLVM 15 by disabling opaque pointers

khiner:
This is a temporary approach until we [migrate to opaque pointers](https://llvm.org/docs/OpaquePointers.html#migration-instructions), and works for me on clang 15.0.5. The only issue I ran into was a new mutex error on shutdown when destructing `ExecutionEngine`: `recursive_mutex lock failed: Invalid argument`.

I don't have sources for llvm in my debug environment (just headers) but I would assume it's in [this line](https://github.com/llvm/llvm-project/blob/main/llvm/lib/ExecutionEngine/ExecutionEngine.cpp#L228). I can't find any info about this API being updated in 15, and couldn't find a resolution. Simply not explicitly deleting the `ExecutionEngine` instance works for me without apparent issues, but it's clearly not good to just remove the delete. I don't see any relevant differences between LLVM's JIT example for [15.0.5](https://github.com/llvm/llvm-project/blob/llvmorg-15.0.5/llvm/examples/HowToUseJIT/HowToUseJIT.cpp) compared with say [14.0.0](https://github.com/llvm/llvm-project/blob/llvmorg-14.0.0/llvm/examples/HowToUseJIT/HowToUseJIT.cpp). They both expect the caller to explicitly `delete`. So something else is going on here.

Anyway, I thought I'd get the ball rolling, and hoped someone with more LLVM/Faust experience could help, especially someone who's building LLVM locally already and can step into it.

### DIFF
--- a/compiler/generator/llvm/llvm_code_container.cpp
+++ b/compiler/generator/llvm/llvm_code_container.cpp
@@ -52,6 +52,9 @@ CodeContainer* LLVMCodeContainer::createScalarContainer(const string& name, int 
 LLVMCodeContainer::LLVMCodeContainer(const string& name, int numInputs, int numOutputs)
 {
     LLVMContext* context = new LLVMContext();
+#if LLVM_VERSION_MAJOR == 15
+     context->setOpaquePointers(false);
+#endif
     Module* module = new Module(gGlobal->printCompilationOptions1() + ", v" + string(FAUSTVERSION), *context);
     
     init(name, numInputs, numOutputs, module, context);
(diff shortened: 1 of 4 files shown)


sletz:
My understanding is that is that LLVM 15 is an intermediate step in the pointer transition process. So 
`setOpaquePointers(true)` seems to be a temporary solution.

sletz:
Thanks. Manually merged with some more macOS improvements in https://github.com/grame-cncm/faust/commit/89f0798ce8470866e1eecd4931fad71bbe386a3e.

PR #838, opened 2022-12-04
WIP Handle opaque pointers for LLVM 15+

khiner:
@sletz Basically trying to translate the recommendations in LLVM's [migration guide](https://llvm.org/docs/OpaquePointers.html#migration-instructions).

Took me a while to even get here, but I'm starting to get the idea after lots of stepping through code. I think this is basically the right approach but could use some help at this point. I put notes in the relevant part where I'm getting stuck, hit me up with any questions! Hoping this start at least helps. I'm probably missing something obvious. I'm invested now, so would love to know how you go about it if you do take it on from here. Otherwise if you have pointers... pun accidental, I'm happy to keep pushing later. Finals week coming up, but after that.

### DIFF
--- a/compiler/generator/llvm/llvm_code_container.cpp
+++ b/compiler/generator/llvm/llvm_code_container.cpp
@@ -52,9 +52,6 @@ CodeContainer* LLVMCodeContainer::createScalarContainer(const string& name, int 
 LLVMCodeContainer::LLVMCodeContainer(const string& name, int numInputs, int numOutputs)
 {
     LLVMContext* context = new LLVMContext();
-#if LLVM_VERSION_MAJOR == 15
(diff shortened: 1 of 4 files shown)


sletz:
Thanks for the try. My personal local version was a bit more involved, but I could neither make it complete. I guess time is now for a bit of general cleanup and restructuration of this LLVM pointer related code. 
But not ASAP, since your temporary version can live until LLVM 16 becomes official, and I have more urgent things todo on my side.

khiner:
Yes it looks like that may be needed to totally switch to opaque pointers. Sounds good on waiting. I’ll close this PR, we can always reference it later. Thanks for looking!

PR #844, opened 2022-12-24
build: patch libsndfile linkage

### LINKED ISSUE #840: 2.54.9: Compiling sound2faust with static libsndfile is a problem for distributions
Since https://github.com/grame-cncm/faust/commit/d1fe8570d2d77839359c0cd6c13bb4abc0c7efb5 sound2faust wants to be compiled with libsndfile.a, but this breaks distribution packaging, which relies on linking against the shared library.

chenrui333:
fixes #840
relates to https://github.com/Homebrew/homebrew-core/pull/118532

### DIFF
--- a/tools/sound2faust/Makefile
+++ b/tools/sound2faust/Makefile
@@ -1,5 +1,5 @@
 
-DESTDIR ?= 
+DESTDIR ?=
 PREFIX ?= /usr/local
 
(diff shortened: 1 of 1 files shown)


sletz:
Fixed in https://github.com/grame-cncm/faust/commit/ca013457c9d52bdc0101c9d31fc3621fe3e1b103

PR #847, opened 2022-12-27
Fix missing `std::` namespace qualifiers in `global.cpp`

khiner:
Likely missed in https://github.com/grame-cncm/faust/commit/fd39eb6375914db8746d9fa9c0410c6e3601abff


sletz:
Fixed in  https://github.com/grame-cncm/faust/commit/6819e539dc36c1472f578d575b71ce8041ea6ab4

PR #850, opened 2022-12-30
Make all compute input buffers const

khiner:
I am in the process of converting my audio backend from [libsoundio](https://github.com/andrewrk/libsoundio) to [miniaudio](https://github.com/mackron/miniaudio), and ran into an issue where miniaudio's main audio callback function provides a `const *input` buffer. To pass this buffer to Faust, I would then have to copy the contents of the const buffer into my own mutable input buffer. (Or maybe use `const_cast`, which is undesirable.) Are there any current use cases where directly modifying the contents of input buffers is desired behavior? If not, I think it would be a nice API improvement to make all `compute` input buffers as `const`.

This PR was generated by replacing all occurrences of `FAUSTFLOAT** input` with `const FAUSTFLOAT** input`. All I did was:
```shell
find -E . -regex '.*\.(c|h|cc|hh|cpp|hpp|md)' -print0 | xargs -0 sed -i '' -e 's/FAUSTFLOAT\*\*\ input/const\ FAUSTFLOAT\*\*\ input/g'
```

and committed.

### DIFF
--- a/architecture/c-jack-gtk.c
+++ b/architecture/c-jack-gtk.c
@@ -165,7 +165,7 @@ class Cdsp : public dsp {
             metadatamydsp(&glue);
         }
     
-        virtual void compute(int count, FAUSTFLOAT** input, FAUSTFLOAT** output)
+        virtual void compute(int count, const FAUSTFLOAT** input, FAUSTFLOAT** output)
         {
             computemydsp(fDSP, count, input, output);
         }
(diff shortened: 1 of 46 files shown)


sletz:
The `-inpl `option allows to change the inputs buffer in place. So I don't think we can do that.

khiner:
Ah gotcha, ok then I'll close this. Thanks!

PR #862, opened 2023-03-02
add libfaust with llvm github action

DBraun:
This sets up a new workflow that builds libfaust dynamic libraries with LLVM on macos, ubuntu, and windows. The workflow only launches via a button on the GitHub actions page. The build results in artifacts which can be downloaded at the bottom of the actions page:

https://github.com/DBraun/faust/actions/runs/4311010847

Windows artifact contains:
* faust.dll
* faust.lib
* faust.exp

macos artifact contains:
* libfaust.2.dylib
* libOSCFaust.a

Note that the Ubuntu zip has duplicate files. To keep it smaller, I need to decide which of the three files to keep:
* libfaust.so <-- I think this is a symbolic link
* libfaust.so.2 <-- probably also symbolic
* libfaust.so.2.58.9

The action/upload-artifact doesn't work well with symbolic links (https://github.com/actions/upload-artifact/issues/93)

I ended up just uploading the entire build/lib folder, and I think that hardened the symbolic links into duplicate files, bloating the ubuntu zip.


sletz:
build-macos  is for x86_64 only, it is possible to build also the ARM version?

DBraun:
> build-macos is for x86_64 only, it is possible to build also the ARM version?

Unfortunately not yet on GitHub actions. Maybe one day they'll have ARM runner images. If it were possible, then the build-macos workflow would be modified like this:

```
          - name: macos-arm64
            os: macos-11
            cmake-options: >-
              -DCMAKE_BUILD_TYPE=Release
              -DCMAKE_VERBOSE_MAKEFILE=ON
              -DCMAKE_OSX_ARCHITECTURES="arm64"
              -DCMAKE_OSX_DEPLOYMENT_TARGET=10.15
            llvm-options: >-
              -DLLVM_DEFAULT_TARGET_TRIPLE="x86_64-apple-darwin19.6.0"
              -DLLVM_TARGETS_TO_BUILD="AArch64"
              -DLLVM_ENABLE_ZLIB=off
              -DLLVM_OPTIMIZED_TABLEGEN=ON
              -DCMAKE_INSTALL_PREFIX="./llvm"
```

sletz:
It seems https://github.com/agraef/pd-faustgen is able to do it, see; 
https://github.com/agraef/pd-faustgen/blob/faustgen2/.github/workflows/makefile.yml
Could it help ?

DBraun:
No more bloat in the ubuntu zip. It's just
* libfaust.so
* libOSCFaust.a

So I think we're good, except still no macos arm automation.

DBraun:
Oh I'll look into what you said about pd-faustgen.

DBraun:
For macos and ubuntu, it looks like libsndfile.a/so are already in the faust repo. Are those suitable for building with libfaust? Do they enable mp3/flac/ogg etc?

sletz:
libsndfile.a/so are not directly need in libfaust, but in architectures files that implement the `soundfile` primitive and addition sound handling tools.

DBraun:
What do you think about including it anyway? If yes, should we build libsndfile as a dynamic or static library?

DBraun:
Let's leave out sndfile. I'm going to have my arm64 mac this weekend and I'll prepare two llvm builds the same way pd-faustgen did.

DBraun:
Closing in favor of #864

PR #877, opened 2023-04-06
Plug intervals

aherrou:
Plug the intervals library into the compiler.


sletz:
Merger in --squash mode with some more cleanup. Thanks !

PR #879, opened 2023-04-13
Corrects the display of intervals in the signal representation

aherrou:
(e.g. when using `faust2sig`)
For one, the opening square bracket [ was converted to its ASCII code 91. Second, when building a type from an interval, the interval's LSB was not transmitted as the resolution of the type.


sletz:
Thanks, merged in --squash mode here: https://github.com/grame-cncm/faust/commit/190402671be44bbfa291c1803cbb1b47286b7cad

PR #883, opened 2023-04-17
Adding submodule for faust2ck pointing to Stanford-CCRMA

celestebetancur:


sletz:
Thank, but I suggest also fixing the https://github.com/ccrma/faust2ck/blob/main/README which is full of old and even non working links.

sletz:
And https://github.com/ccrma/faust2ck/blob/main/src/chuck_faust.template.cpp is also quite old and should be updated. I'll prepare a PR.

celestebetancur:
Thank for updating the template, There is a small thing that I am keeping for backward compatibility from the older version. Also the readme is also updated.

PR #884, opened 2023-04-18
Github Action w/ LLVM updates

DBraun:
Major changes to Github Actions. The goal is to build everything statically on the GitHub action but allow users to use dynamic libraries when building locally, such as libsndfile, gtk etc. For faustgen, and various tools we also link statically against libfaustwithllvm.


sletz:
Merged in https://github.com/grame-cncm/faust/commit/9ecbf81fb16fe640012c6e9cd0739c3bb46f1ac5.

PR #888, opened 2023-04-22
Faust2supercollider removal

olafklingt:
remove faust2supercollider because it fails silently because faust2sc doesn't exist anymore


sletz:
I don't understand. `faust2supercollider` is still used in some use-cases. What is thisPR about ?

olafklingt:
faust2supercollider depends on faust2sc which doesn't exist anymore. the UGens generated with faust2supercollider don't get a proper .sc file generated that describes the UGen.

sletz:
Are you speaking of https://github.com/grame-cncm/faust/tree/master-dev/tools/faust2sc-1.0.0 ?

olafklingt:
ah yes sorry i didn't see it

PR #889, opened 2023-04-23
Set repr(C) on the Dsp struct to recover performance

joshtriplett:
After https://github.com/rust-lang/rust/pull/102750 , Rust reorders
fields in repr(Rust) structures based on size/alignment. Applying such
reordering to the Dsp struct substantially hurts performance.

Set repr(C) on the Dsp struct to recover that performance.

See
https://internals.rust-lang.org/t/unexpected-3-x-performance-regression-starting-with-rust-version-1-67/18724
for the test case and investigations that led to this.


sletz:
Thanks.

PR #892, opened 2023-05-04
Removes deprecated method that fails daisy build

### LINKED ISSUE #868: Error with -pod option in faust2daisy
Hi, 
I'm facing a little problem with faust2daisy
with -pod option it throw this error 
ex_faust.cpp: In function 'int main()':
ex_faust.cpp:10446:8: error: 'class daisy::DaisyPod' has no member named 'Configure'
10446 |     hw.Configure();
      |        ^~~~~~~~~

Without the -pod option the compilation is successful but I don't have any sound

Best

colinsullivan:
When using the `-patch` and `-pod` arguments, the build will fail because the `Configure` method has been deprecated in the latest Daisy SDK.

Closes #868


sletz:
Thanks.

PR #894, opened 2023-05-11
Draft release action

DBraun:
I can run the libfaust action with the checkbox for making a draft release and it resulted in a draft release with all of the artifacts.


sletz:
Thanks.

Issue #899, opened 2023-05-15
rm: cannot remove "preset.cpp": No such file or directory


magnetophon:
When I run a lot of faust2jack instances in parallel, I sometimes get the above error, presumably coming from https://github.com/grame-cncm/faust/blob/2.59.6/tools/faust2appls/faust2jack#L158 or the faust2lv2 equivalent.

I discovered this when I added a Makefile to [Voice of Faust](https://github.com/magnetophon/VoiceOfFaust).
When I run ``make -j24`` this issue turns up.
When I run ``make -j1`` it does not.

Could it be that when there are 2 or more faust2jack scripts running, they sometimes delete files that don't belong to them?

Is there anything I can do to help fix this?

PS: This is with version 2.59.6.
I *think* 2.59.5 did not have this issue, but that doesn't make sense, since as far as I can tell nothing changed between those versions that could be related.

PR #900, opened 2023-05-16
make preset management thread safe

magnetophon:
Fixes https://github.com/grame-cncm/faust/issues/899


sletz:
StepGPT checked, thanks!

PR #904, opened 2023-05-20
use XDG_DOCUMENTS_DIR as the default preset dir

magnetophon:
This is still a work in progress in the sense that only ``faust2jaqt`` is implemented.
The helper functions should probably not live in ``jack-qt.cpp``, since they will be the same for all architectures that use presets.
I hope this version is better then https://github.com/grame-cncm/faust/pull/903!

## Reasoning

There are two changes:

A) The presets don't go in ``PRESETDIR/name_preset0.000000`` anymore, but instead go in ``PRESETDIR/name/preset0.000000``.
This is a bit more organized, if you ask me.

B) PRESETDIR is chosen a bit more flexible now.
These are the steps the algorithm takes, and my reasoning for each one. 

1) If PRESETDIR is ``auto``, it tries to create a directory using the ``XDG_DOCUMENTS_DIR`` environment variable.
If I am making presets, I probably want to keep them.
In the previous implementation, the presets could be gone after a reboot.
``$XDG_DOCUMENTS_DIR`` is the closest we have to [a standard](https://wiki.archlinux.org/title/XDG_user_directories#Creating_custom_directories) location for files the user wants to keep.

2) If PRESETDIR is a valid existing directory, it returns the path with a trailing slash. 
This is almost the same as in the current implementation: we don't try to create the directory but we just use it.
The difference is that we check if it is a valid directory and if not, go on to the next option.

3) If PRESETDIR is an environment variable, it tries to create a directory using its value. 
This is so you can change the preset directory at runtime.
For example:
```bash
export MY_PRESET_DIR=/home/me/presets
faust2jaqt -preset MY_PRESET_DIR test.dsp
```
We now use ``/home/me/presets/test``
```bash
export MY_PRESET_DIR=/home/me/presets2
./test
```
We now use ``/home/me/presets2/test``
Note we didn't need to recompile for this.
```bash
faust2jaqt -preset $MY_PRESET_DIR test.dsp
```
Note the ``$``; we are now stuck in ``/home/me/presets2/test``.

4) If ``$HOME`` is set, it tries to create a directory at HOME/Documents or HOME/PRESETDIR. 
HOME/Documents in the default value for XDG_DOCUMENTS_DIR, so it seems a good next fallback.
HOME/PRESETDIR is a bit weird, but at least it is in the home dir.

5) If all else fails, it falls back to /var/tmp/.
This is the directory the current implementation uses for ``auto``.

### DIFF
--- a/architecture/faust/gui/PresetUI.h
+++ b/architecture/faust/gui/PresetUI.h
@@ -27,6 +27,17 @@
 
 #include <string>
 #include <iostream>
+#include <cstdlib>
+#include <cstring>
+#include <climits>
+
+#if __cplusplus < 201703L // If the version of C++ is less than 17
+#include <experimental/filesystem>
+namespace fs = std::experimental::filesystem;
+#else
+#include <filesystem>
+namespace fs = std::filesystem;
+#endif
 
 #include "faust/gui/DecoratorUI.h"
 #include "faust/gui/FUI.h"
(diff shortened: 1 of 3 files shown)


sletz:
All preset management code should better go in [PresetUI.h](https://github.com/grame-cncm/faust/blob/master-dev/architecture/faust/gui/PresetUI.h), this was it will be naturally shared.

sletz:
I simplified the code in https://github.com/grame-cncm/faust/commit/eb39d083e022fb35d2068d2241cf38ae0aff0202 and merged the branch. Can you check that is still behaves correctly in all cases ? Thanks.

PR #905, opened 2023-05-21
Refctor faustgen_factory class to separate files

DBraun:
This refactors the `faustgen_factory` class into separate .h/.cpp files. There should be no functional changes. Ideally we wouldn't have `#include "maxcpp5.h"` in `faustgen_factory.h` because I wanted it to be agnostic to Max, but I didn't untangle it that far yet.


sletz:
Small fix to make it work, thanks.

PR #906, opened 2023-05-25
MapUI: remove redundant map lookups

jatinchowdhury18:
The `MapUI.h` architecture file has a few spots where it needs to get/set a parameter value from a `std::map<std::string, FAUSTFLOAT*>`. Previously this was being done with the pattern:
```cpp
if (fPathZoneMap.find(str) != fPathZoneMap.end()) {
    return *fPathZoneMap[str];
}
```
With this approach, both `fPathZoneMap.find(str)` and `fPathZoneMap[str]` are re-doing the same work of finding the relevant entry in the map, which is a $\mathcal{O}(\text{log}(n))$ operation in the size of the map.

This PR re-writes that pattern as
```cpp
const auto fPathZoneMapIter = fPathZoneMap.find(str);
if (fPathZoneMapIter != fPathZoneMap.end()) {
    return *fPathZoneMapIter->second;
}
```
So the work of finding the entry in the map only happens once.

### DIFF
--- a/architecture/faust/gui/MapUI.h
+++ b/architecture/faust/gui/MapUI.h
@@ -149,19 +149,19 @@ class FAUST_API MapUI : public UI, public PathBuilder
          */
         void setParamValue(const std::string& str, FAUSTFLOAT value)
         {
-            auto fPathZoneMapIter = fPathZoneMap.find(str);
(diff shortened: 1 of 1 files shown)


sletz:
Indeed, same issue at other places in the code ((-;, but in less critical code path AFAICS. 
Thanks !

PR #923, opened 2023-07-09
JuceAssertWhenUsingTabs

jdsierral:
When trying a faust structure with tabs within tabs I hit an assert in juce_TabbedButtonBar.cpp in line 248. This is because the added tab didn't have a name since it was created with openTabBox which in contrast to openVerticalBox or openHorizontalBox doesn't have a label assigned to it.

Fixed this by adding setName to the constructor of uiTabBox but maybe there is a better way to do this since all the other constructors are empty and these type of things are done through member initializers.

Hope this helps.
JuanS


sletz:
Thanks !

PR #925, opened 2023-07-18
Use repr(C) by default

Gui-Yom:
Follow up to https://github.com/grame-cncm/faust/issues/875#issuecomment-1640221245


sletz:
Thanks !

bluenote10:
Thanks for the fix and also the main refactoring! From what I've seen so far this is a great improvement!

PR #933, opened 2023-08-06
Ground work for a VHDL backend

AloisRautureau:
Hello,

This PR includes the implementation of a VHDL backend for the FAUST compiler.
Said backend is separated into two distinct parts: one implementing VHDL-specific optimization passes on signals, and another generating the actual source code.

The former is complete, while the latter misses some key elements (operators, a good implementation of delays, user-interface signals, etc).

Regards,
Aloïs RAUTUREAU


sletz:
"This branch cannot be rebased due to conflicts " can you rebase your branch first ?

PR #936, opened 2023-08-20
Fix build error, and allow multiple calls to `DSPToBoxes` without resetting lib context.

khiner:
`createDSPFactoryFromBoxes` already resets the global context (inside `boxesToSignalsAux`), but only a single call to `DSPToBoxes` will currently be successful until the client calls `destroy/createLibContext`.

### DIFF
--- a/architecture/faust/dsp/libfaust-signal.h
+++ b/architecture/faust/dsp/libfaust-signal.h
@@ -133,7 +133,7 @@ struct Interval {
     {}
 };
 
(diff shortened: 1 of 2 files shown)


sletz:
Thanks, I actually called `reset` at several places in https://github.com/grame-cncm/faust/commit/953d5cfce243e7d5fa424cbba824cf3ea1c2b6ad

khiner:
Thanks! I almost added those cases as well but wasn't sure about it.

PR #941, opened 2023-08-29
change realPropagate(BoxRoute) to use simplifyingAdd() rather than si…

oleg-nesterov:
random test-case:

    o_t = 2,1 : route(2,2,1,2,2,1);
    one = o_t : _,!;
    two = o_t : !,_;
    process = route(two,two,one,two,two,one);

### DIFF
--- a/compiler/normalize/aterm.cpp
+++ b/compiler/normalize/aterm.cpp
@@ -47,7 +47,7 @@ aterm::aterm(Tree t)
 /**
  * Add two terms trying to simplify the result
  */
-static Tree simplifyingAdd(Tree t1, Tree t2)
+Tree simplifyingAdd(Tree t1, Tree t2)
 {
     faustassert(t1);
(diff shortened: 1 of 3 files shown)


sletz:
Thanks. Instead of exporting and using `simplifyingAdd`, I  prefer using `simplify` which is already used at several places in `propagate`.

Done in: https://github.com/grame-cncm/faust/commit/428eee11896556707b78f9044d66d2397c9884fa.

sletz:
The way I did here: https://github.com/grame-cncm/faust/commit/428eee11896556707b78f9044d66d2397c9884fa#diff-8abe84c88df958389ad964a0de5e0f01f24b1b8057b004faf5aa5a89a5149c10R540
breaks some DSPs. But the simpler ` simplifyingAdd(exp, lsig[src - 1]);` change the generated code in a way that I still don't understand. So reverting for now.

PR #947, opened 2023-09-27
Update faust2jackconsole to allow effects option

akjmicro:
The `faust2jackconsole` script didn't have an effects option. Fixing. Also, fixing compile error b/c `-ldl` was missing from the linker options.


josmithiii (commented):
Looks good to me

PR #965, opened 2023-10-29
Update README.md

bhargavshirin:
In the sentence "A distinctive characteristic of Faust is to be fully compiled," it might be clearer to say "A distinctive characteristic of Faust is that it is fully compiled."


sletz:
Thanks, part of the PR added manually.

PR #977, opened 2023-11-25
change BinOp("|") to have fLeftAbsorbing/fRightAbsorbing predicates

oleg-nesterov:
Before this patch

	import("stdfaust.lib");
	process = |(ma.not(0));

compiles to

	output0[i0] = FAUSTFLOAT(int(float(input0[i0])) | -1);

after the patch

	output0[i0] = FAUSTFLOAT(-1);


sletz:
Merged thanks.

Issue #986, opened 2024-01-02
mesh2faust failing to build on Ubuntu 22.04 (Win 11 WSL)


retromore:
Trying to compile mesh2faust on Ubuntu 22.04 (Win 11 WSL);
After fixing all dependencies the compiler gets to 97% and then throws the following error:

```
[ 97%] Building CXX object CMakeFiles/mesh2faust-bin.dir/src/main.cpp.o
/home/rtt/mesh2faust/src/main.cpp: In function ‘int main(int, char**)’:
/home/rtt/mesh2faust/src/main.cpp:180:33: error: no matching function for call to ‘mesh2faust(const char*&, <brace-enclosed initializer list>)’
  180 |     string dsp = m2f::mesh2faust(
      |                  ~~~~~~~~~~~~~~~^
  181 |         objectFileName,
      |         ~~~~~~~~~~~~~~~
  182 |         {
      |         ~
  183 |             modelName,
      |             ~~~~~~~~~~
  184 |             materialProperties,
      |             ~~~~~~~~~~~~~~~~~~~
  185 |             freqControl,
      |             ~~~~~~~~~~~~
  186 |             modesMinFreq,
      |             ~~~~~~~~~~~~~
  187 |             modesMaxFreq,
      |             ~~~~~~~~~~~~~
  188 |             targetNModes,
      |             ~~~~~~~~~~~~~
  189 |             femNModes,
      |             ~~~~~~~~~~
  190 |             exPos,
      |             ~~~~~~
  191 |             nExPos,
      |             ~~~~~~~
  192 |             showFreqs,
      |             ~~~~~~~~~~
  193 |             debugMode
      |             ~~~~~~~~~
  194 |         }
      |         ~
  195 |     );
      |     ~
In file included from /home/rtt/mesh2faust/src/main.cpp:1:
/home/rtt/mesh2faust/src/mesh2faust.h:34:13: note: candidate: ‘std::string m2f::mesh2faust(const char*, m2f::MaterialProperties, m2f::CommonArguments)’
   34 | std::string mesh2faust(
      |             ^~~~~~~~~~
/home/rtt/mesh2faust/src/mesh2faust.h:36:24: note:   no known conversion for argument 2 from ‘<brace-enclosed initializer list>’ to ‘m2f::MaterialPropertie’
   36 |     MaterialProperties materialProperties = {}, // material properties to set on the generated volumetric mesh
[...]

PR #987, opened 2024-01-02
mesh2faust: Fix out-of-order arg in main.

khiner:
Also fixes Eigen dirs not being included in executable target. (It was only included with library).

Fixes https://github.com/grame-cncm/faust/issues/986


sletz:
Thanks!

PR #1006, opened 2024-03-30
iOS Check SoundFile Bundle Path Condition in LibsndfileReader

RuolunWeng:


sletz:
Thanks. AFAICS your way of doing also work on other OS, so better use it everywhere with this commit: https://github.com/grame-cncm/faust/commit/c756e428547a725bdb4f76738eab055ffeb29a41

And the error message is not displayed because we don't want to pollute the console here.

PR #870, opened 2023-03-24
Mesh2faust eigen

khiner:
cc @sletz @rmichon @josmithiii

## Goal

This PR is about making `mesh2faust` work on non-Intel processors (such as M1 Macs) by removing the dependency on MKL. With this PR, I have successfully created `.lib` files for Romain's example bell `.obj` files [linked here](https://ccrma.stanford.edu/~rmichon/pmFaust/) on Ubuntu 20.10/x86_64 and MacOS Ventura/M1 Pro.

## Details

The only thing that MKL is being used for solving the sparse eigenvalue problem. After looking at the landscape I decided to go with Eigen and [Spectra](https://github.com/yixuan/spectra) instead of Pardiso/MKL for finding the eigenvalues/eigenvectors. This seems like a reasonable approach since the library is robust, focused on only this problem, header-only, and its only dependency is Eigen. I added Spectra as a submodule and directed users to install the Eigen library on their system using a package manager in the readme.

This PR removes all BLAS/LAPACK/Pardiso/MKL setup & dependency. Eigen doesn't depend on any of these, using its algorithms if none are linked. However, it does automatically defer to BLAS/LAPACK/MKL in some circumstances if one is linked (see [here](https://eigen.tuxfamily.org/dox/TopicUsingBlasLapack.html) for details). I also removed Vega's solver lib and all solver headers (except `CGSolver.h`, which is used in an example file).

## Runtime

Here is a gist showing a runtime comparison, along with a 1:1 comparison of the old vs new eigenvalues:
https://gist.github.com/khiner/c97788198bbc69b04d718128acdfcb34

I found a lot of variance in the runtimes, but this is pretty representative - it basically tends to take around the same amount of time.

## Not great things

One thing I don't like about this PR is that, since I haven't replaced the FEM parts of the code, it still relies on Vega and uses Vega's `SparseMatrix` classes in addition to the new `Eigen::SparseMatrix` class. This is just a minor annoyance, though, since it trades a bit of explicit conversion for getting rid of a whole bunch of more hidden baggage with the Vega/Pardiso/MKL chain.

Other annoying things are Eigen's default column-major ordering, and Spectra's largest-first ordering, which is different than Pardiso/Vega's smallest-first ordering. (This is more sensible, but it means a bit more index juggling.)

Thank you for taking a look! I'm happy to implement suggested changes etc., to get this to a point everyone's happy with.

PR #1019, opened 2024-05-06
mesh2faust improvements

khiner:
Improvements to `mesh2faust` tool:
* Add damping estimation to the Finite Element model, following [this paper](https://www.cs.cornell.edu/~djames/papers/DyRT.pdf)
* Fixed a bad shift value being passed to the eigenvalue solver, which was causing poor frequency estimates for low-frequency modes.
* Return estimated mode details in API result (frequencies/decay times/per-vertex mode gains in new `ModalModel` struct) in addition to the DSP code string.
* Split the code-gen phase into a separate function and other cleanups.

Some results can be found [here](https://github.com/khiner/MeshEditor).

cc @rmichon @josmithiii ([not for review](https://github.com/grame-cncm/faust/pull/870#issuecomment-1713305568), just so you're aware)


sletz:
Thanks.

PR #1024, opened 2024-05-25
Large refactor to allow for "helper" scripting

bassmanitram:
* Move the architecture stuff to folder 'chaos-stratus'
* Implement a set of tools for connecting to and executing commands on the pedal
* Adapt faust2stratus to use those tools

Eventually the faustservice Makefiles for the Stratus will also use those tools


sletz:
I see `_pedal-tools.sh ` , is the underscore at the beginning a typo ?

bassmanitram:
> I see `_pedal-tools.sh ` , is the underscore at the beginning a typo ?

No - its a hint that people shouldn't try to use it as a script - fairly common practice chez nous :)

PR #1048, opened 2024-08-09
Updated shebang in shell scripts

Yoann-Le-Borgne:
Hi!

Some scripts were using a `#!/bin/sh` shebang with _bash_ specific syntax instead of a POSIX compliant one, which is called a bashism.
Historically, _sh_ has been linked to _dash_ in Ubuntu and Debian (see https://wiki.ubuntu.com/DashAsBinSh). This is fine most of the time but there are cases when this can lead to errors.
Using _checkbashisms_ from the _devscripts_ package we can see that this is the case for some shell scripts (you can run `grep -r '#!/bin/sh' -l | xargs checkbashisms` at the root of the project to see by yourself).
As most of the scripts already use `#!/bin/bash`, this PR modifies the shebang for the remaining ones in the _tools_ folder

Regards,
Yoann


sletz:
Thanks.

PR #1056, opened 2024-10-03
rust const definitions should use usize

crop2000:
currently printMacros declares in rust constants that have the type i32 but those variables represent vector/array length as such their usage is usually at places where usize is expected. additionally they cannot be negative.


sletz:
Thanks, can you just correct the typos in the PR title and push again in --force mode?

sletz:
"usize is _used used_ in rust for array/vector sizes"  another try ?

sletz:
Thanks.

Issue #1059, opened 2024-10-05
removing trait impl from rust generator


crop2000:
The current [rust code generation](https://github.com/grame-cncm/faust/tree/master-dev/compiler/generator/rust) has some limitations. The functions associated with the generated struct are wrapped into one trait. The trait defines the interface, but the interface might depend on compile options. For example [-ec](https://github.com/grame-cncm/faust/issues/1057) should be implemented for rust as well. It would break the current trait-interface. So one would need multiple traits. 

But I think the code generation should not define any trait. Rather, the traits should be defined and implemented in the architecture files. I thought about architecture files primarily as files that provide usability scaffolding like [here](https://github.com/Frando/rust-faust/blob/main/faust-build/faust-template.rs). I didn't expect single file programs like in the faust repository. I hope it is enough to provide backward compatible architecture files, like this for the [faust-build crate](https://github.com/crop2000/rust-faust/blob/try_less_traits/faust-build/faust-template.rs).

The change would reduce the complexity of the code generated by rust and give more duties to the architecture files.

I am currently preparing a pr regarding this issue.

sletz:
The general strategy we have for the C/C++ backend and architecture model is needed to put less in the code generation step, and more in the architecture files, since this is more flexible in general. The current Rust code generation has ben developed first in 2017, with little Rust experience at that time, and refine a bit from there. So yes this kind of change is welcome. 

The only thing to be sure is to provide backward compatibility, and possibly have feedback from @obsoleszenz, @plule...etc. form [this contributor list](https://github.com/Frando/rust-faust/graphs/contributors)

crop2000:
There is a limit to backwards compatibility.
If people have a program that has a reference to the trait FaustDsp (not the dsp struct) then the trait implementation is needed and the used architecture file must to be updated. 
This is the case in rust-faust/rust-state crate.
I will update the architecture files rust-faust and faust, for as a reference.

crop2000:
I would favor a cleanup which reduces the generated code to the most necessary ... and provides const rather then functions for get_num_inputs for example.  

but this will be definitely not backward compatible. 

now i am divided between 2 thoughts (but still favor solution 1).

1) because people need to add  the FaustDsp trait implementation to their own architecture files anyway, we should change the generated code to be as simple as possible.

2) because people might use their own architecture files we should keep as much of the current interface in the dsp struct impl so that they only need to remove references to FaustDsp from their code which works well if they haven't depend on the trait anyway.

[...]

PR #1062, opened 2024-10-09
Rust code generation independent of FaustDsp trait

crop2000:
faust code generation should not specify traits.
this pr introduces this change.
It is tested against the impulse-tests using the architecture file in its subfolder.
How much architecture files need to change depends on how traits are used.

The jack cpal and minimal architecture files could be compiled after minimal changes, 
but need more cleanup work.

### DIFF
--- a/compiler/generator/rust/rust_code_container.cpp
+++ b/compiler/generator/rust/rust_code_container.cpp
@@ -122,7 +122,7 @@ void RustCodeContainer::produceInternal()
     // generateInstanceInitFun("instanceInit" + fKlassName, false, false)->accept(&fCodeProducer);
 
     tab(n + 1, *fOut);
-    *fOut << "fn instance_init" << fKlassName << "(&mut self, sample_rate: i32) {";
+    *fOut << "pub fn instance_init" << fKlassName << "(&mut self, sample_rate: i32) {";
     tab(n + 2, *fOut);
     fCodeProducer.Tab(n + 2);
     generateInit(&fCodeProducer);
@@ -136,12 +136,12 @@ void RustCodeContainer::produceInternal()
     string counter = "count";
     if (fSubContainerType == kInt) {
         tab(n + 1, *fOut);
-        *fOut << "fn fill" << fKlassName
+        *fOut << "pub fn fill" << fKlassName
               << subst("(&mut self, $0: i32, table: &mut[i32]) {", counter);
     } else {
         tab(n + 1, *fOut);
-        *fOut << "fn fill" << fKlassName
-              << subst("(&mut self, $0: i32, table: &mut[$1]) {", counter, ifloat());
+        *fOut << "pub fn fill" << fKlassName
+              << subst("(&mut self, $0: i32, table: &mut[FaustFloat]) {", counter);
     }
     tab(n + 2, *fOut);
     fCodeProducer.Tab(n + 2);
@@ -168,9 +168,68 @@ void RustCodeContainer::produceInternal()
     *fOut << "}";
 }
 
+void RustCodeContainer::produceFaustDspBlob()
+{
+    const char*  tab  = "  ";
+    *fOut << "impl FaustDsp for " << fKlassName << " {" << endl;
+    *fOut << tab << "type T = FaustFloat;" << endl;
+    *fOut << tab << "fn new() -> Self where Self: Sized {" << endl;
+    *fOut << tab << tab << "Self::new()" << endl;
+    *fOut << tab << "}" << endl;
+    *fOut << tab << "fn metadata(&self, m: &mut dyn Meta) {" << endl;
+    *fOut << tab << tab << "self.metadata(m)" << endl;
+    *fOut << tab << "}" << endl;
+    *fOut << tab << "fn get_sample_rate(&self) -> i32 {" << endl;
+    *fOut << tab << tab << "self.get_sample_rate()" << endl;
+    *fOut << tab << "}" << endl;
+    *fOut << tab << "fn get_num_inputs(&self) -> i32 {" << endl;
+    *fOut << tab << tab << "FAUST_INPUTS as i32" << endl;
+    *fOut << tab << "}" << endl;
+    *fOut << tab << "fn get_num_outputs(&self) -> i32 {" << endl;
+    *fOut << tab << tab << "FAUST_OUTPUTS as i32" << endl;
+    *fOut << tab << "}" << endl;
+    *fOut << tab << "fn class_init(sample_rate: i32) where Self: Sized {" << endl;
+    *fOut << tab << tab << "Self::class_init(sample_rate);" << endl;
+    *fOut << tab << "}" << endl;
+    *fOut << tab << "fn instance_reset_params(&mut self) {" << endl;
+    *fOut << tab << tab << "self.instance_reset_params()" << endl;
+    *fOut << tab << "}" << endl;
+    *fOut << tab << "fn instance_clear(&mut self) {" << endl;
+    *fOut << tab << tab << "self.instance_clear()" << endl;
+    *fOut << tab << "}" << endl;
+    *fOut << tab << "fn instance_constants(&mut self, sample_rate: i32) {" << endl;
+    *fOut << tab << tab << "self.instance_constants(sample_rate)" << endl;
+    *fOut << tab << "}" << endl;
+    *fOut << tab << "fn instance_init(&mut self, sample_rate: i32) {" << endl;
+    *fOut << tab << tab << "self.instance_init(sample_rate)" << endl;
+    *fOut << tab << "}" << endl;
+    *fOut << tab << "fn init(&mut self, sample_rate: i32) {" << endl;
+    *fOut << tab << tab << "self.init(sample_rate)" << endl;
+    *fOut << tab << "}" << endl;
+    *fOut << tab << "fn build_user_interface(&self, ui_interface: &mut dyn UI<Self::T>) {" << endl;
+    *fOut << tab << tab << "self.build_user_interface(ui_interface)" << endl;
+    *fOut << tab << "}" << endl;
+    *fOut << tab << "fn build_user_interface_static(ui_interface: &mut dyn UI<Self::T>) where Self: Sized {" << endl;
+    *fOut << tab << tab << "Self::build_user_interface_static(ui_interface);" << endl;
+    *fOut << tab << "}" << endl;
+    *fOut << tab << "fn get_param(&self, param: ParamIndex) -> Option<Self::T> {" << endl;
+    *fOut << tab << tab << "self.get_param(param)" << endl;
+    *fOut << tab << "}" << endl;
+    *fOut << tab << "fn set_param(&mut self, param: ParamIndex, value: Self::T) {" << endl;
+    *fOut << tab << tab << "self.set_param(param, value)" << endl;
+    *fOut << tab << "}" << endl;
+    *fOut << tab << "fn compute(&mut self, count: i32, inputs: &[&[Self::T]], outputs: &mut [&mut [Self::T]]) {" << endl;
+    *fOut << tab << tab << "self.compute(count as usize, inputs, outputs)" << endl;
+    *fOut << tab << "}" << endl;
+    *fOut << "}" << endl;
+}
+
 void RustCodeContainer::produceClass()
 {
     int n = 0;
+    tab(n, *fOut);
+    *fOut << "pub type FaustFloat = " << ifloat() << ";";
+
     tab(n, *fOut);
     *fOut << "use std::convert::TryInto;";
 
@@ -292,16 +351,7 @@ void RustCodeContainer::produceClass()
 
     tab(n, *fOut);
     *fOut << "impl " << fKlassName << " {";
-    generateCompute(n + 1);
     tab(n, *fOut);
-    *fOut << "}" << endl;
-    tab(n, *fOut);
-
-    *fOut << "impl FaustDsp for " << fKlassName << " {";
-
-    // Associated type
-    tab(n + 1, *fOut);
-    *fOut << "type T = " << ifloat() << ";";
 
     // Memory methods
     tab(n + 2, *fOut);
@@ -326,7 +376,7 @@ void RustCodeContainer::produceClass()
         tab(n + 1, *fOut);
     }
 
-    *fOut << "fn new() -> " << fKlassName << " { ";
+    *fOut << "pub fn new() -> " << fKlassName << " { ";
     if (fAllocateInstructions->fCode.size() > 0) {
         tab(n + 2, *fOut);
         *fOut << "allocate" << fKlassName << "(dsp);";
@@ -349,10 +399,6 @@ void RustCodeContainer::produceClass()
     tab(n + 1, *fOut);
     *fOut << "fn get_sample_rate(&self) -> i32 { self.fSampleRate as i32}";
     tab(n + 1, *fOut);
-    *fOut << "fn get_num_inputs(&self) -> i32 { FAUST_INPUTS as i32}";
-    tab(n + 1, *fOut);
-    *fOut << "fn get_num_outputs(&self) -> i32 { FAUST_OUTPUTS as i32}";
-    tab(n + 1, *fOut);
 
     // Inits
 
@@ -364,7 +410,7 @@ void RustCodeContainer::produceClass()
     // generateInstanceInitFun("instanceInit" + fKlassName, false, false)->accept(&codeproducer2);
 
     tab(n + 1, *fOut);
-    *fOut << "fn class_init(sample_rate: i32) {";
+    *fOut << "pub fn class_init(sample_rate: i32) {";
     {
         tab(n + 2, *fOut);
         // Local visitor here to avoid DSP object type wrong generation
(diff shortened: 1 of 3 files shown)


sletz:
Thanks. Testing here with [faust2portaudiorust](https://github.com/grame-cncm/faust/tree/master-dev/architecture/rust#faust2portaudiorust) tool, but I gets runtime error like: 

```
Faust Rust code running with Portaudio: sample-rate = 44100 buffer-size = 64
get_num_inputs: 1
get_num_outputs: 2
thread 'main' panicked at src/main.rs:227:11:
called `Result::unwrap()` on an `Err` value: InvalidChannelCount
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

so I guess the [current port audio architecture](https://github.com/grame-cncm/faust/blob/master-dev/architecture/rust/portaudio-float.rs),  which use a fixed `const CHANNELS: i32 = 2;`. 

Can you have a look ?

crop2000:
ah now i understand the context of the other comment on the other pr better ... yes i will have a look.

bluenote10:
Could we wait a little bit with making this change? I have a substantial 'productive' usage of Rust + Faust, and I would like to test how this change would impact it. Unfortunately I'm travelling right now, and will only have access to a PC in a few days again.

Specifically I want to verify if using ambiguous method vs impl names doesn't cause issues (hopefully it doesn't require using fully qualified names). Also, I'm seeing that some of the trait methods have been removed, and I'm wondering if it is easy to work around the missing functionality.

sletz:
The merged commit is here: https://github.com/grame-cncm/faust/commit/28a8b1a9218057dd5cfb3a3a736bd03c614375c3 and is not involving any change in the architectures files. So hopefully it should not break anything.  Please yes give feedback. 

In a more general term, @crop2000 we definitively have to be quite careful. As I said, I don't have expertise and time enough to be the final judge. @bluenote10, @obsoleszenz at least have to give their feedback each time.

obsoleszenz:
> Could we wait a little bit with making this change? I have a substantial 'productive' usage of Rust + Faust, and I would like to test how this change would impact it. Unfortunately I'm travelling right now, and will only have access to a PC in a few days again.
> 
> Specifically I want to verify if using ambiguous method vs impl names doesn't cause issues (hopefully it doesn't require using fully qualified names). Also, I'm seeing that some of the trait methods have been removed, and I'm wondering if it is easy to work around the missing functionality.

Agree. Let's move this one carefully. Maybe first introduce it behind a feature flag. But for sure am with @crop2000 that the rust code gen could be improved a good bit.

crop2000:
@sletz yes commit 28a8... does not change the interface at all. It extents it providing a new function.

@bluenote10  when you are able to share your architecture file it would be interesting. I don't aim to rush this PR, but i like to keep it minimal so i don't want to improve the architecture files in this PR. Just want to keep them working in the context they worked before. (Outdated jack etc...)

crop2000:
@obsoleszenz a feature flag would increase the maintainance burden. Because it would double the code.

I rather try to find a solution that works with a drop in extension, the reproduces the old behavior.

bluenote10:
> The merged commit is here: https://github.com/grame-cncm/faust/commit/28a8b1a9218057dd5cfb3a3a736bd03c614375c3 and is not involving any change in the architectures files.

I can confirm that 28a8b1a9218057dd5cfb3a3a736bd03c614375c3 works well for me, and I like the idea of having that `compute_array` function.

> when you are able to share your architecture file it would be interesting. 

My architecture file is rather boring:

```rust
use crate::types::{FaustDsp, Meta, UI, ParamIndex};

type F32 = f32;

// Generated intrinsics:
<<includeIntrinsic>>

// Generated class:
<<includeclass>>
```

What is crucial in my use case it to have a common interface for all generated DSPs (I have many), i.e., I naturally need a trait, and all generated DSPs are supposed to re-use the same base types.

I have experimented a little bit with this feature branch, and I was able to make my usages work with it as well (by also forwarding the missing methods). So in principle I would be fine with the change.

However, I don't really see yet why we want to go in that direction. So far I see a few minor downsides:

- Now that DSPs are not implemented against an interface out-of-the-box, new developers may wonder how to arrive at a "common denominator" for the generated output (wasn't the original code generation non trait based, and we introduced traits to make the set of methods more obvious?). In other words, it is not obvious that forwarding the various `self.xxx` calls in the architecture file into a trait is generally valid, because it is not clear whether one can rely on a particular `self.xxx` being generated.
- For use cases that want a trait, the forwarding of all method calls into a trait makes the generated code larger: Every dsp will contain the regular `impl` block plus the trait forwarding boilerplate. Not a big problem, but avoiding the boilerplate part would be nicer.
- When using the same method in the traits as the underlying inherent function names, one may start wondering whether the ambiguity can cause issues (worst case would be an infinite recursion, if e.g. `self.compute` would call the trait method instead of the inherent function). Most likely this should not be an issue, because fortunately Rust behaves as desired in this case, but it creates at least unnecessary questions. Also note that there seems to be a clippy lint (https://github.com/rust-lang/rust-clippy/issues/11499) that warns about such ambiguities. Using _different_ method names in the trait to avoid the ambiguity is also not ideal due to reduced readability.

If I understand it correctly, this change was motivated by supporting `-ec` which (probably) requires different signatures e.g. of the `compute` method. What would be the implications of keeping the `FaustDsp` trait, but treating its method signatures as "variable"? I.e., if invoked with `-ec` switch the codegen would expect the `FaustDsp` trait (injected via the architecture file) to look slightly different. This would probably imply that one has to use different architecture files depending on the command line switches. Is that what you are trying to avoid?

crop2000:
Hi @bluenote10,
Thanks for your reply.

My motivation for this and following changes is to keep the code generated by faust small and at the same time make the interface more flexible.

I consider the FaustDsp trait flawed. It seems to me that it is designed like a object. It has for example the anti pattern to specify a new function for the trait. 

The functions I removed from the trait 

can never be used as trait functions one always need to refer to the specific object that needs to implement the specific struct (mydsp by default) because they don't have a self argument.

Did you try to generate your code with those functions removed from the trait specification you import?

crop2000:
@bluenote10 I have an idea. 

we move this block https://github.com/grame-cncm/faust/blob/2026d7a11e97552ce44539936884cc6635e4fde8/tests/impulse-tests/archs/rust/architecture.rs#L246 
into the code generated by faust (with the 3 functions i criticized above).
and add a flag like --no-faustdsp-trait
to make it optional.

This allows us to be fully backward compatible.
And have a separation of concerns (code generated by faust, traits, backward compatibility).
And we can change the underlying implementations without affecting current architecture files.

In the long run I think we can come up with traits that are better specified ... the FaustDsp implementation would in that case redirect to those implementations.

This doesn't solve the issue with the clippy lint you mentioned but this lint would appear in code generated by rust and not in the architecture file so i think this would be less of a problem.

bluenote10:
> My motivation for this and following changes is to keep the code generated by faust small 

For use cases that need a trait the generated code is significantly larger due to having two impl blocks now. My git stats on my project after incorporating this change says ~500 lines of code more.

> I consider the FaustDsp trait flawed. It seems to me that it is designed like a object.

That is rather vague. The trait is not object safe, it is supposed to be used statically / generically.

> It has for example the anti pattern to specify a new function for the trait.

To clarify: Having static factory methods in a trait is per se obviously not an anti-pattern -- it's basically just what the `Default` trait does. You are probably referring to not using `Default` to begin with? The idea here was to eventually change the signature to `new(sample_rate: u32) -> Self` to avoid always to set the sample rate after the construction (which is madatory before usage anyway). If the static `new` method has a signature different from `Default` (because DSPs cannot really be default constructed without specifying the sample rate) then the anti-pattern is gone.

> The functions I removed from the trait cannot be called from a trait object, because they don't have a self argument.

Well, of course, they are static factory methods, so you have to use them like e.g. `MyDspType::build_user_interface_static(ui_interface)` or via generics.

> Did you try to generate your code with those functions removed from your trait specification?

Yes, but then everything breaks for me. My usages are all generic based on the trait. Think of usages like:

```rust
pub fn create<D: FaustDsp<T = AudioFloat>>(sample_rate: i32) -> D {
    let mut instance = D::new();
    instance.init(sample_rate);
    instance
}

pub fn extract_param_defs<D: FaustDsp<T = AudioFloat>>() -> Vec<DspParamDef>
where
    T: Clone,
{
    let mut ui_builder = DeclarativeUIBuilder::new();
    D::build_user_interface_static(&mut ui_builder);
    ui_builder.get_param_defs()
}
```

----------------

For the record, I have pulled out the history where we switched from non-trait to trait: https://github.com/grame-cncm/faust/pull/448 At that time, this was generally preferred by e.g. @bkfox as well (https://github.com/grame-cncm/faust/issues/409#issuecomment-644698703), because their use case also required having a common interface for all generated DSPs.

I think it basically comes down to:

> My motivation for this and following changes is to [...] make the interface more flexible.

The question is _how_ flexible does it have to be? The flip side of flexibility is instability and accidental breaking changes.

For usages that need a common denominator for multiple generated DSPs arbitrary flexibility can be detrimental: Each method generated inside the impl block would raise the question: Is this method guaranteed to exist in all DSPs? Is this method signature always the same (e.g. `compute_arrays` already isn't). Implementing the code gen against an interface/trait brings clarity to these questions, because it documents _these are the methods you can expect to get from the code generation_. If the code generation would fail to generate a method, generates additional methods, or generates different signatures, the Rust compilation would fail with a clear error message. Without a trait it would be much easier that a contributor to Faust opens PR that removes/modifies certain function signatures without noticing, which are breaking changes to Rust users. So far it was relatively easy for me to keep track of such breaking changes, because I could simply look at modifications to the trait in the architecture file, which served as a documentation of what to expect from the code generation output.

> did you see this discussion: https://github.com/grame-cncm/faust/issues/1059

Yes, I wasn't really sure where to comment. Feel free to move the discussion back to that issue.

bluenote10:
> we move this block [...] into the code generated by faust (with the 3 functions i criticized above).

That is how it currently works, right? https://github.com/grame-cncm/faust/blob/eaead8383df13351786d92b2f4a9390f9955a81a/compiler/generator/rust/rust_code_container.cpp#L289

> and add a flag like --no-faustdsp-trait to make it optional.

This could work, but doesn't it make the generation rather complex because it needs to support placing the method in either impl blocks?

I would love to understand how much flexibility we really need to make up my mind if giving up on traits is really better.

architecture/rust/jack-float.rs
```diff
@@ -150,10 +150,10 @@ fn main() {
         let output0: &mut[f32] = &mut out_a_p;
         let output1: &mut[f32] = &mut out_b_p;
 
-        let inputs = &[input0, input1];
-        let outputs = &mut[output0, output1];
+        let inputs = vec![input0, input1];
+        let mut outputs = vec![output0, output1];
```
obsoleszenz, line 154: Why use a vec here? This is probably a heap allocation, even if it's just a small one.

obsoleszenz:
I find it difficult to really see what this pr changes, can we have some before/after code generation example?

crop2000:
@obsoleszenz 

https://gist.github.com/crop2000/9f6a0c628df5c920c6f83e3ea5df5540 as generated in the impulse-tests using https://github.com/crop2000/faust/blob/rust_no_traits/tests/impulse-tests/archs/rust/architecture.rs

please ignore the changes to the other architecture files for now.

crop2000:
@bluenote10 
when you want to use MyDspType::build_user_interface_static(ui_interface) or new() why not implement it on the struct?

crop2000:
one point about what i think is important about traits they should only define what is needed for one task. In our case the information for UI and the compute function are not directly related and should have separate traits.

It would be very bad to let the generated code change the FaustDsp trait itself 
so that all code that expects FaustDsp following a certain definition would fail. 

For example by including compute_arrays() to FaustDsp 
or to add control()  when -ec flag is used to generate the code) 

the latest proposal to keep the FaustDsp as an optional feature in the generated code helps to ensures backward compatibility as long as the trait definition is not changed.

crop2000:
in commit [e21f3f0](https://github.com/grame-cncm/faust/pull/1062/commits/e21f3f051082ef5c6727e9e8866ec5f81724748c) i moved the FaustDsp trait impl block into a function in code generation for backwards compatibility.

I see that this makes sense as a reference implementation to ensure backwards compatibility. 
I like this idea because it would free us to change the underlying implementation more freely. 

for example if compiled with -ec flag the implemetation for compute in FaustDsp would either panic or call self.control()  before self.compute() ... just as an example that it creates space for improvements and behavior changes.

architecture/rust/jack-float.rs
crop2000: it was a leftover change, sorry

compiler/generator/rust/rust_code_container.cpp
```diff
@@ -204,13 +266,13 @@ void RustCodeContainer::produceClass()
         tab(n, *fOut);
         *fOut << "}";
         tab(n, *fOut);
-        *fOut << "fn remainder_f32(from: f32, to: f32) -> f32 {";
+        *fOut << "pub fn remainder_f32(from: f32, to: f32) -> f32 {";
```
bluenote10, line 269: Why make these (`remainder_f32`, `rint_32`, `remainder_64`, `rint_f64`) public?

As far as I can see they are only supposed to be used privately so they can be kept private, no?

compiler/generator/rust/rust_code_container.cpp
```diff
@@ -549,20 +601,20 @@ void RustCodeContainer::generateComputeHeader(int n, std::ostream* fOut,int fNum
     // Compute "compute" declaration
     tab(n, *fOut);
     tab(n, *fOut);
-    *fOut << "fn compute_arrays("
+    *fOut << "pub fn compute_arrays("
```
bluenote10, line 604: I find it a bit strange that in the generated code the current order of methods is:

- `compute_arrays` comes first.
- then comes `new`, and all other "basic" methods, followed by UI related methods.
- the last method is `compute`.

Since `compute` and `compute_arrays` have a direct relationship, it would be nice to generate them next to each other. Could we just move these generating lines down into `generateComputeInterfaceHeader`?

compiler/generator/rust/rust_instructions.hh
```diff
@@ -287,7 +287,7 @@ class RustInstVisitor : public TextInstVisitor {
             // impls, we need a mechanism to forward the information whether to use "pub"
             // or not. In the worst case, we have to prefix the name string like "pub fname",
             // and handle the prefix here.
-            *fOut << "fn " << inst->fName;
+            *fOut << "pub fn " << inst->fName;
```
bluenote10, line 290: What is this change about? (I couldn't really figure out what it affects during the code gen.)

Is the comment actually still valid after this change, because now function are no longer only attached to a trait as the first sentence says?

bluenote10 (approved):
This change looks generally good now after the last adaptation. Forwarding from the trait methods to the inherent methods is probably a viable strategy and there is no breaking change.

compiler/generator/rust/rust_instructions.hh
crop2000: i think it influenced the generation of get_num_inputs() before i changed it in the other pr.
i would remove the pub again.
Since functions are not attached it wouldn't matter if they are prefixed with "pub".
But because this is for internal functions we don't need pub.
I don't think the comment makes sense so I would remove it completely.

compiler/global.cpp
```diff
@@ -451,7 +451,9 @@ void global::reset()
     gGroupTaskSwitch = false;
     gFunTaskSwitch   = false;
 
-    gUIMacroSwitch = false;
+    gUIMacroSwitch    = false;
+    gRustNoTaitSwitch = false;
```
bluenote10, line 455: Nitpick: `gRustNoTaitSwitch` -> `gRustNoTraitSwitch`?

documentation/man/README.md
```diff
@@ -284,3 +286,4 @@ AUTHOR
 Copyright (C) 2002-2024, GRAME - Centre National de Creation Musicale.
 All rights reserved.
 
+make[1]: Leaving directory '/home/olaf/projects/rust/forks/faust/documentation/man'
```
bluenote10, line 289: Committing this line looks a bit like an accident?

crop2000: I guess i should not run this make step myself i guess it should be done in the versiinig step

crop2000:
I am thinking to remove the option to not generate the trait from this pr. Maybe a slightly different option (Generating less code) would be desirable but having to many options is not. Also changing the interface twice is not.

compiler/global.cpp
crop2000: thanks, fixed but this commit got removed from this pr.

crop2000:
@bluenote10 do you agree with this pr?
Before FaustDsp was the implementation of the generated Faust code.
Now FaustDsp is a interface to the generated Faust code that allows us to change the underlying implementation, while keeping backwards compatibility.

bluenote10:
> @bluenote10 do you agree with this pr?

Yes, as far as I can see this strategy should be fine.

It probably makes sense to merge #1066 first, because this PR seems to (partially?) contain the same changes, right?

crop2000:
this pr contains the commit of #1066 because it builds on it. I opened a separate pr to keep the scope of each change clear.

crop2000:
@sletz this PR is now also ready to be merged.
I removed all changes to architecture files because for the impulse-tests this proves that this pr is backward compatible.

sletz:
Merged thanks. Be sure to check `faust2portaudiorust` and `faust2jackrust` tools. I had to update architecture files here: https://github.com/grame-cncm/faust/commit/0dbcf51f933595622fed7cb565de9a1c2736aed8

crop2000:
I will review the architecture files in the next days

PR #1065, opened 2024-10-12
add unity visionOS compatibility and fix unity utilities JSON parser to handle unexpected values

tactevo:
The utility code generated by faust2unity fails to parse the most recent changes to embedded JSON metadata, particularly the "size" key, which should be ignored.   This patch skips unknown JSON metadata keys as long as they are double-quoted string values or numbers.

In addition, I've added basic compatibility for VisionOS (ie, macro checks now allow for UNITY_VISIONOS) which result in VisionOS projects using ios libs.


sletz:
Thanks.

PR #1066, opened 2024-10-14
Always provide const values for rust

crop2000:
This commit removes the -uim flag for rust
and provides the const values by default
and makes use of these const values
in get_num_inputs and get_num_outputs
it also simplifies how this functions are generated they are now just a string

To create arrays for compute_arrays() one needs to know the array sizes at compile time. The const definitions provide these.

### DIFF
--- a/compiler/generator/code_container.cpp
+++ b/compiler/generator/code_container.cpp
@@ -801,15 +801,6 @@ void CodeContainer::printMacros(ostream& fout, int n)
                 tab(n, fout);
             }
             fout << "#endif" << endl;
-        } else if (gGlobal->gOutputLang == "rust") {
-            fout << "pub const FAUST_INPUTS: usize = " << fNumInputs << ";";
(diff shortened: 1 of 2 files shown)


crop2000:
@bluenote10 I am quite sure that this will not break anything. but can you have a look.

bluenote10 (approved):
This looks good to me, and I can confirm that there is no breaking change:

Nitpick: I would have generated these `pub const FAUST_XXX` expressions somewhere at the top of the file instead of the very bottom, because such constants are traditionally placed more in the "header" area of a file (and the other PR also places the `FaustFloat` alias on top, which plays a similar role).

crop2000:
I moved the const definition just before the struct definition.
I think this pr is ready for merging

crop2000:
@sletz are you open to merge this?

sletz:
Thanks.

Issue #1060, opened 2024-10-06
rust impulse test fails on linux 


crop2000:
faust at commit fea73ffdd90ed4ee978933c5967c7ac872f60f21 (HEAD -> master-dev)
rustc 1.78.0 (9b00956e5 2024-04-29) (built from a source tarball)
Linux 6.1.108-rt40 https://github.com/grame-cncm/faust/pull/1-NixOS SMP PREEMPT_RT Wed Sep 4 11:25:05 UTC 2024 x86_64 GNU/Linux

```
crop@yoga ~/p/r/f/f/t/impulse-tests (master-dev)> make rust
make -f Make.rust FAUSTOPTIONS="-I dsp -double"
make[1]: Entering directory '/home/crop/projects/rust/forks/faust/tests/impulse-tests'
mkdir -p ir/rust
../../build/bin/faust -lang rust -I dsp -double -i -A ../../architecture -a archs/rust/architecture.rs dsp/vcf_wah_pedals.dsp -o archs/rust/src/bin/vcf_wah_pedals.rs
#cd archs/rust/ && cargo run --bin vcf_wah_pedals ../../ir/rust/vcf_wah_pedals.ir
cd archs/rust/ && cargo run --release --bin vcf_wah_pedals ../../ir/rust/vcf_wah_pedals.ir
   Compiling faust-impulse-test v0.1.0 (/home/crop/projects/rust/forks/faust/tests/impulse-tests/archs/rust)
error[E0308]: mismatched types
    --> src/bin/vcf_wah_pedals.rs:3205:91
     |
3205 | ...0 {if iSlow28 != 0 {(fTemp73 * fTemp56 / fTemp60 + self.fRec69[1] * fTemp61 + 2.0 * (fTemp75 * self.fRec69[0] / fTemp58) + (1.0 - fTemp51 - 2.0 * fTemp54 * fTemp75) * fRec70 / (fTemp58 * F64::sqrt(fTemp55))) / fTemp50} else {(self.fRec67[2] + fTemp47 + self.fRec67[0]) / fTemp44}} e...
[...]

PR #1067, opened 2024-10-21
Fix regression because of  #1061

crop2000:
solve unimportant warning 
and 
and solve regression because of fix for #1060.
adding osc_enable.ir to rust impulse tests revealed a mistake in the fix for #1060
This PR removes brackets around if when it is not from a select2 instruction.


bluenote10:
Are the Rust impulse tests not executed by the CI (GitHub Action) before merging a PR?

sletz:
No, I still do manual tests, and this regression was not seen locally.

crop2000:
no and the test that showed the problem wasn't part of the test set.

PR #1068, opened 2024-10-21
Attempt to run impulse tests in CI

bluenote10:
Would it be possible to enable the impulse tests in the CI? It would add some confidence during the review process to immediately see that they all pass. Or are there any deeper reasons why the tests have to be executed manually only?


bluenote10:
I didn't had time yet to investigate the test failures, but fortunately @crop2000 already did in #1073! So I guess best to continue there.

PR #1073, opened 2024-10-26
2b) add cpp and rust tests on ci 

crop2000:
based on #1068 
a lot of work was caused by #1075

### DIFF
--- a/.github/workflows/ubuntu.yml
+++ b/.github/workflows/ubuntu.yml
@@ -2,7 +2,7 @@ name: Ubuntu
 
 on:
   push:
-    branches: 
+    branches:
         - '*'
   pull_request:
     branches: [ master-dev ]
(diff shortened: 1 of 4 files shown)


.github/workflows/ubuntu.yml
```diff
@@ -13,11 +13,29 @@ jobs:
...
+         mkdir -p faust/dsp/
+         cp compiler/generator/libfaust.h faust/dsp/
+
+    - name: Install Rust
+      uses: dtolnay/rust-toolchain@stable
+      with:
+        toolchain: stable
+
+    - name: Run impulse tests (C++)
+      run: cd tests/impulse-tests && make travis
```
bluenote10, line 38: Strange that the target is called "travis" (tests don't run on Travis CI, no?). But of course that's somewhat independent of this PR, if it happens to be called like this already.

crop2000: It is a subset of the cpp tests that finishes in below 1h

crop2000:
@sletz what do you think about this? 
To run the tests takes about 80min.

sletz:
This is too long, I suggest modifying the `make travis` target to only keep the first line.

crop2000:
@sletz is the travis target used elsewhere? if not i would rename it to github_action and describe a meaning full subset. maybe it makes sense to have a detailed check that is not part of the default actions, but can be activated to ensure a thorough check for more complex changes.

crop2000:
@sletz down to ~20m for everything ~5min build faust, ~15min c++ and rust

PR #1072, opened 2024-10-26
3) implement ec and cm flag for rust

crop2000:
implement and test ec and cm flag for rust

PR #1076, opened 2024-10-27
4) rust -os flag

crop2000:


crop2000:
based on top of #1077 #1080 #1072

crop2000:
@sletz finally this passes the tests too, so it is ready.

compiler/generator/instructions_compiler.cpp
```diff
@@ -1138,9 +1138,14 @@ ValueInst* InstructionsCompiler::generateVariableStore(Tree sig, ValueInst* exp)
                     return IB::genLoadStructVar(vname);
                 }
             } else {
-                // "Slow" variables are declared as locals in 'frame' or 'compute' functions
-                pushComputeBlockMethod(IB::genDecStackVar(vname, ctype, exp));
-                return IB::genLoadStackVar(vname);
+                if (gGlobal->gOneSample) {
```
sletz, line 1141: Why this change is needed ?

compiler/generator/instructions_compiler.cpp
```diff
@@ -1178,9 +1183,13 @@ ValueInst* InstructionsCompiler::generateVariableStore(Tree sig, ValueInst* exp)
                             getConditionCode(sig), IB::genStoreStructVar(vname_perm, exp)));
                         return IB::genLoadStructVar(vname_perm);
                     } else {
-                        // copy the object variable to the local one
-                        pushComputeBlockMethod(
-                            IB::genDecStackVar(vname, ctype, IB::genLoadStructVar(vname_perm)));
+                        if (gGlobal->gOneSample) {
```
sletz, line 1186: Why this change is needed ?

compiler/generator/rust/rust_code_container.cpp
```diff
@@ -632,13 +636,56 @@ void RustCodeContainer::generateComputeHeader(int n, std::ostream* fOut, int fNu
...
           << "&mut self, " << fFullCount << ": usize, inputs: & [& [FaustFloat] ]"
           << ", outputs: & mut[& mut[FaustFloat] ]) {";
     tab(n + 1, *fOut);
 }
 
+void RustCodeContainer::generateComputeFrame(int n)
+{
+    tab(n, *fOut);
+    tab(n, *fOut);
+    *fOut << "pub fn compute_frame(&mut self, inputs: &[&FaustFloat], ";
```
sletz, line 649: Can you use `frame` instead of `compute_frame` to be similar to what C and C++ already generate ?

compiler/generator/instructions_compiler.cpp
sletz: This break the C/C++ backends. Compare the generated code in Rust and C/C++.

crop2000: when i -os and don't have -ec i use the generateControlDeclarations before the "loop" part. only when i put "slow" variable into the declare block they will be available. same for the other comment

sletz: Change in `Instructions_compiler.cpp` may break other backends which is the case here. So try to save the issue in the Rest backend itself and do not change `Instructions_compiler.cpp`.

compiler/generator/rust/rust_code_container.cpp
crop2000: i prefer compute_frame because it makes clearer that it is a variant of compute.

crop2000:
@sletz i changed this pr and also tested cpp with -os in ci

sletz:
Thanks, merged, with the "public" function renaming and different fix in the tests (make c/make cpp were broken).

PR #1077, opened 2024-11-03
1) improve impulse-test makefile for rust

crop2000:
The aim of this pr is to make the Makefile for rust tests more concise similar to the Make.gcc .
The motivation is to also test  osc_enable.dsp where applicable. 
Unlike that Makefile i also test for error messages.
Any comments?
@sletz 
@bluenote10


sletz:
Merged thanks.

PR #1079, opened 2024-11-07
Implement SDF3 generation backend

jkmingwen:
Add implementation for SDF3 generation from Faust programs.

With this implementation, SDF3 graphs (in the form of XML files) can be generated using the `-lang sdf3` flag, and the output file can be specified with `-o <file>`. Currently included in the default, "most", and "all" builds in the Makefile.


sletz:
Thanks. I did somme improvements, and merged in https://github.com/grame-cncm/faust/commit/8289f4f93a79310222d87007860c305768481fa8

PR #1080, opened 2024-11-10
2) Make Rust FaustDsp Trait Optional

crop2000:
This PR introduces the new compile flag `-rnt`  `--rust-no-faustdsp-trait` to faust, to not generate the FaustDsp interface.
The aim is to be able to make use of flags like `-ec` or `-os` that change the generated code so that it is not compatible with the FaustDsp trait. 
@bluenote10 
@obsoleszenz

### DIFF
--- a/tests/impulse-tests/Makefile
+++ b/tests/impulse-tests/Makefile
@@ -120,6 +120,12 @@ travis:
 	#$(MAKE) -f Make.gcc outdir=cpp/double/sch     lang=cpp arch=impulsearch.cpp FAUSTOPTIONS="-I dsp -double -sch"
 
 #########################################################################
+# automatic github action test
+github_action:
+	$(MAKE) -f Make.gcc outdir=cpp/double          lang=cpp arch=impulsearch.cpp FAUSTOPTIONS="-I ../../libraries/ -I dsp -double"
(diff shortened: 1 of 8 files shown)


tests/impulse-tests/Makefile
```diff
@@ -362,13 +362,13 @@ interp1:
 #########################################################################
 # Rust backend
 rust:
-	$(MAKE) -f Make.rust FAUSTOPTIONS="-I dsp -double"
+	$(MAKE) -f Make.rust FAUSTOPTIONS="-I dsp -double -rnt"
```
bluenote10, line 365: In my opinion the CI should always use the mode that generates code against the trait.

As explained in the old PR, this largely reduces the risk of making an accidental breaking change. If the CI no longer runs against the trait based generation it would become rather easy to open a PR and merge it, and by accident breaks the interface in a fundamental way.

Personally I would prefer if `-ec` and `-os` would also use _their own_ traits instead to clearly communicate what the stable interface of these modes is.

crop2000: Ok i will setup different combinations of flags for the tests. 

I don't want to work on faust generating traits now because i think traits are a separate problem to solve. People who want a simple way of using faust can just use less flags.

crop2000:
rebased on top of #1077

sletz:
Merged, thanks.

PR #1087, opened 2024-11-25
5) rust: use AsRef/AsMut in compute interface

crop2000:
the current compute function signature looks like this:
```
	pub fn compute(&mut self, count: usize, inputs: &[&[FaustFloat]], outputs: &mut [&mut [FaustFloat]]){
```
this prevents us from passing a mutable reference as input buffer slice.
in the case one wants to chain dsps where the outputs of one dsp are used as the input of another it would be good if one can pass pass same slice that is used as the output of one dsp as input of another. but in that case one would pass a mutable reference. which is not possible with the current function signature.
this pr changes the functions to be more general. so that one can pass a mutable reference to the the input.
The new now looks like this:
```
	pub fn compute<I, O>(
		&mut self,
		count: usize,
		inputs: &[I],
		outputs: &mut [O],
	) where
		I: AsRef<[FaustFloat]>,
		O: AsMut<[FaustFloat]>,
	{
```
this change should be fully backwards compatible. (as first attemps have shown that needs some review and improvements of testing)

Because I found out that my earlier introduced compute_array interface is not contributing to the performance i removed it here again because it would make this change more complicated.

### DIFF
--- a/tests/impulse-tests/Makefile
+++ b/tests/impulse-tests/Makefile
@@ -122,10 +122,9 @@ travis:
 #########################################################################
 # automatic github action test
 github_action:
-	$(MAKE) -f Make.rust outdir=rust/osec CARGOOPTIONS="" FAUSTOPTIONS="-I ../../libraries/ -double -os -ec -rnt -a archs/rust/architecture_osecrnt.rs"
-	$(MAKE) -f Make.rust outdir=rust/os CARGOOPTIONS="" FAUSTOPTIONS="-I ../../libraries/ -double -os -rnt -a archs/rust/architecture_osrnt.rs"
 	$(MAKE) -f Make.rust outdir=rust/no CARGOOPTIONS="" FAUSTOPTIONS="-I ../../libraries/ -double -a archs/rust/architecture_trait.rs"
-	$(MAKE) -f Make.gcc outdir=cpp/double/os lang=cpp arch=impulsearch7ter.cpp FAUSTOPTIONS="-I dsp -double -os"
+	$(MAKE) -f Make.rust outdir=rust/cm CARGOOPTIONS="" FAUSTOPTIONS="-I ../../libraries/ -double -cm -a archs/rust/architecture_cm.rs"
+	$(MAKE) -f Make.rust outdir=rust/ecrnt CARGOOPTIONS="" FAUSTOPTIONS="-I ../../libraries/ -double -ec -rnt -a archs/rust/architecture_ecrnt.rs"
 	$(MAKE) -f Make.gcc outdir=cpp/double lang=cpp arch=impulsearch.cpp FAUSTOPTIONS="-I ../../libraries/ -I dsp -double"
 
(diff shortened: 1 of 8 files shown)


crop2000:
@sletz this one is also ready

tests/impulse-tests/Makefile
```diff
@@ -122,10 +122,7 @@ travis:
 #########################################################################
 # automatic github action test
 github_action:
-	$(MAKE) -f Make.rust outdir=rust/osec CARGOOPTIONS="" FAUSTOPTIONS="-I ../../libraries/ -double -os -ec -rnt -a archs/rust/architecture_osecrnt.rs"
```
sletz, line 125: Why do you remove the tests here ? (I think it make sense to keep all of them)

crop2000: The rational was to only keep tests that relate to a change in ci. I don't have a strong opinion on this.

sletz:
You wrote **this change is fully backwards compatible** but `faust2portaudiorust` and `faust2jackrust` are broken after this change.  So what are existing projects supposed to do ?

crop2000:
I made a mistake before because i thought i already implemented the most general interface.
Now i wrote a test to compare different interfaces and updated the pr accordingly.
[here](https://gist.github.com/crop2000/12af81bd2093fb43cacf6f90ef7af90d) is the test for documentation.
I also updated some architecture files so that the cover differently typed buffers to be passed to the compute function.

crop2000:
I think i will update the PR one more time to match the function signature as i updated it in the pr description above.

bluenote10:
>  @bluenote10 could I ask your "more expert than me" overview on this PR ?

Sure I can look into this in a bit more detail, but will probably only get to it next weekend.

crop2000:
i just realized the function signature can be simplified with the same meaning:
```
pub fn compute(&mut self, count: usize, inputs: &[impl AsRef<[FaustFloat]>], outputs: &mut [impl  AsMut<[FaustFloat]>])
```
lets see if i can make this change before the review ... but functionally this should be equivalent to the current version.

bluenote10:
I had a closer look now into this, and as far as I can see this should be all fine!

I was re-rerunning [all my benchmarks](https://github.com/bluenote10/RustFaustExperiments/tree/master/Benchmarks) before and after that change to get a better feeling of how the generated code looks, and if it has an impact on performance. For what I've seen it all looks fine to me, and I would assume that it likely produces the same assembly anyway.

crop2000:
@sletz @bluenote10 thanks for having a look. I am also preparing benchmarks,  but havnt cleaned them up.

sletz:
> @sletz @bluenote10 thanks for having a look. I am also preparing benchmarks, but havnt cleaned them up.

Should I wait for that before merging ?

crop2000:
@sletz no need to wait for publishable benchmarks.

sletz:
Merged, thanks.

PR #1096, opened 2024-12-15
make use of occurrences instead of sharing property

crop2000:
To me it seems that the sharing property is obsolete and superseded by Occurrences.
This pr removes the use of sharing and makes use of occurrences.


sletz:
Thanks, will be needed later on.

crop2000:
What would be needed for this idea to be merged?

PR #1099, opened 2024-12-25
rust: FaustDsp trait improvements

crop2000:
This PR has braking changes to the faust interface via the FaustDsp trait.
I provides the same type signature of compute in FaustDsp as I do for direct use.
I think this simplifies code significantly as vectors as well as slices can be used as arguments.
Additionally i change the type of the return value of get_num_input get_num_output methods. 
The change implies some changes as the change makes FaustDsp lose its  trait-object safty.
faust-state in rust-faust needs to be changed from a dyn box (dynamic dispatch) to a impl (static dispatch) as [done here](https://github.com/crop2000/rust-faust/commit/4d767a09dfa468a90b6aca5fdd7a348efebf7ded#diff-d0ff5010397e73912922507f32118dccd0d21b8d39dec3ad5eecc0411b83ea37R290).

### DIFF
--- a/architecture/rust/cpal.rs
+++ b/architecture/rust/cpal.rs
@@ -58,8 +58,8 @@ pub trait FaustDsp {
     fn new() -> Self where Self: Sized;
     fn metadata(&self, m: &mut dyn Meta);
     fn get_sample_rate(&self) -> i32;
-    fn get_num_inputs(&self) -> i32;
-    fn get_num_outputs(&self) -> i32;
+    fn get_num_inputs(&self) -> usize;
+    fn get_num_outputs(&self) -> usize;
     fn class_init(sample_rate: i32) where Self: Sized;
     fn instance_reset_params(&mut self);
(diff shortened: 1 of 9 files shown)


crop2000:
@bluenote10 can you have a look at this? it is a breaking change but I hope it doesn't make common ways usage impossible.

crop2000:
https://doc.rust-lang.org/reference/items/traits.html#object-safety it seems a lot more can be left as it was if compute is marked Sized.

bluenote10:
> @bluenote10 can you have a look at this? it is a breaking change but I hope it doesn't make common ways usage impossible.

As far as I can see, it is crucial that the trait stays object safe, and that `compute` (and all essential methods) does not use a `where Self: Sized` markers. Otherwise it becomes impossible to use the DSP instances via dynamic dispatch. This would basically break all my use cases.

crop2000:
Yes i agree that's why i closed this pr

Issue #1089, opened 2024-11-26
faust2daisy generated code does not compile for -pod or -patch option


ghoppla:
### faust2daisy generated code does not compile for -pod or -patch option

When compiling code using `faust2daisy` with option -pod or -patch the code does not compile. 
Only -patchsm is working. 

The problem is that somehow ValueConverter.h is not included in the generated cpp file. 

### Steps for reproduction 

**Versions Used:**
- Faust master-dev, 2.76.0, Commit: [5ed7170](https://github.com/grame-cncm/faust/commit/5ed7170ba01bf5f536988d77d90c3235c6ea82fc)
- libDaisy v7.1.0

**Daisy Toolchain Setup on Linux:**

1. Create a folder, e.g.
   ```bash
   $ mkdir ~/daisy
   ```

2. Get arm-none-eabi:
   - See [DaisyWiki](https://github.com/electro-smith/DaisyWiki/wiki/1d.-Installing-the-Toolchain-on-Linux)

3. Get libDaisy:
   ```bash
   $ git clone --recurse-submodules https://github.com/electro-smith/libDaisy
   ```

4. Create `environment.rc` file with the following content:
   ```bash
   export LIBDAISY_DIR=~/daisy/libDaisy

   GCC_PATH=~/daisy/gcc-arm-none-eabi-10-2020-q4-major/bin
   export PATH=$GCC_PATH:$PATH
   ```

5. Compile libDaisy:
   ```bash
   $ source environment.rc
   $ cd libDaisy
   $ make
   ```

**Compilation:**
1. dsp file minimal_example.dsp
```faust
import("stdfaust.lib");
declare name "Daisy Seed Bug";
declare options "[midi:on]";

gate = button("gate");
vel = hslider("gain", 0, 0, 1, .01);
freq = hslider("freq", 440, 0, 20000, 0.01);

process = os.triangle(freq) * vel : _*en.adsr(.1,.1,.7,.2, gate) <: _,_;
```

2. Generated cpp code with:
```bash
$ faust2daisy -patch -midi -sr 48000 -bs 32 -uim -source minimal_example.dsp
```

when running make within the generated folder 
```bash
$ cd minimal_example
$ make
```
a lot of the compilation errors appear, since ValueConverter.h is not included: 

```
ex_faust.cpp:1220:1: error: expected class-name before '{' token
 1220 | {
      | ^
ex_faust.cpp:1265:29: error: 'ValueConverter' was not declared in this scope
[...]

PR #1104, opened 2024-12-29
Daisy and Hothouse pedal updates

DBraun:
### Daisy updates
I merged the `DaisyPatchInitControlUI.h` into `DaisyControlUI.h`, closing https://github.com/grame-cncm/faust/issues/1089. I tested it on a Daisy Pod Rev5 and checked that the two buttons and two knobs are mapped correctly. When `-pod` is not used, I left the code as is for anyone using Daisy Seed on its own or the Daisy patch.Init().

I've added an optional `-sdram` flag so that large buffers like long delay lines can be put in SDRAM. Since it's optional, this is backwards compatible. It uses a Python script `faust_sdram_converter.py`, requiring `python3`.

### Hothouse Pedal
I added a `faust2hothouse` tool and `gui/HothouseControlUI.h` to be used with the [Cleveland Hothouse DIY Pedal](https://clevelandmusicco.com/hothouse-diy-digital-signal-processing-platform-kit/). This script supports UI control for the 6 knobs, 3 togglewitches, and 2 footswitches.


architecture/daisy/README.md
```diff
@@ -2,34 +2,52 @@
...
+The **faust2daisy** tool can be used to program the [patch.Init()](https://electro-smith.com/products/patch-init).
+
+## Bootloader Configuration:
 
 Setup the board to utilize the manufacturer's bootloader, ensuring that the generated code fits within the board's memory. For detailed instructions on flashing the bootloader, please refer to: [Daisy Bootloader Getting Started Guide](https://electro-smith.github.io/libDaisy/md_doc_2md_2__a7___getting-_started-_daisy-_bootloader.html).
 
-## DSP examples
+## Memory Limits (SDRAM)
+
+If you're using a Daisy with significant amount of SDRAM such as the Pod or patch.Init, then you may want to use the `-sdram` flag when compiling code involving long delay lines/tables etc. Enabling this will execute a Python script inside `faust2daisy` that modifies the generated C++ code to put large float buffers (e.g., delay lines) in SDRAM. For this to work, you must have `python3` available on the command line.
```
sletz, line 48: Do you think a custom memory manager could be used here instead of modifying the generate C++ code ? (see https://faustdoc.grame.fr/manual/architectures/#custom-memory-manager)

DBraun: Yea I know that using Python here is not the Faust way, but I'm just trying to get a working solution before an upcoming workshop. We can implement something better later.

sletz: OK

PR #1112, opened 2025-01-20
add varname to Unity template

meta-meta:
I imported a Faust2Unity plugin into my Unity project. I was hitting this error which indicated the JSON for generating UI failed to parse. Stepping through in the debugger, it was failing because there was no case for `varname`.

```cs
if (!FaustUI.fJSONParser(ref fJSON, out fUI)) { // Parses the JSON file
      UnityEngine.Debug.LogError("Error JSON Parser");
}
```


sletz:
Thanks. some similar ones were still there, so everything has been fixed here: https://github.com/grame-cncm/faust/commit/052a69e2d5dfc42bc8a7652dc8a36a44591123f5.

PR #1123, opened 2025-02-20
Add parameter tracking and FFun/FVar in SDF3 backend

jkmingwen:
Updated SDF3 backend to account for FFun and FVar nodes; would previously throw errors when encountered. Also track UI and delay parameter details in SDF actor names.


sletz:
Thanks.

PR #1124, opened 2025-02-23
Add new `mesh2faust` API method to generate modal model and Faust DSP code independently

khiner:
This allows users to compute the modal model separately from the DSP code. It is technically API-breaking since the existing `modal2faust` response changes from a `Response` type to a `std::string`, but I doubt anyone else is using these extra methods. The main intended API is the `mesh2faust` methods, which haven't changed. The only thing missing from the `Response`-typed response is the `ModalModel` they passed into the method anyway.


sletz:
Thanks.

PR #1131, opened 2025-04-03
Add pipewire support with console, qt and gtk GUI.

Pri-yan-shu:
Adds pipewire support, taking inspiration from jack architecture files and faust2jack files.

### DIFF
--- a/architecture/faust/audio/pipewire-dsp.h
+++ b/architecture/faust/audio/pipewire-dsp.h
@@ -64,18 +64,18 @@ class pipewireaudio : public audio {
 
     protected:
         ::dsp* fDSP;
-        pw_thread_loop *loop;
-        pw_filter *filter;
+        pw_thread_loop *fLoop;
+        pw_filter *fFilter;
         std::vector<audio_port*> fInputPorts;
         std::vector<audio_port*> fOutputPorts;
         bool fAutoConnect;
 
-        pw_filter_events filter_events = {
+		// pipewire doesn't take ownership of these two so we have to manage them ourselves
+        pw_filter_events fFilter_events = {
             .version = PW_VERSION_FILTER_EVENTS,
             .process = _pw_process
         };
-
-        spa_hook filter_listener;
+        spa_hook fFilter_listener;
 
         static void _pw_process(void* data, spa_io_position *position)
         {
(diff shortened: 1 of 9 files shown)


sletz:
@Pri-yan-shu AFAICS this PR seems LLM generated, without deep understanding of the code. If you can correct it and prove that is actually works, then fine, otherwise I'll close it.

Pri-yan-shu:
@sletz I have started from jack architecture files and faust2 scripts as base and modified the jack specific api with pipewire. All 3 scripts work and generate perfectly running programs with working midi support. If you want I can send the video demonstration over faust discord.

dromer:
Not super important, but shouldn't the dates in the file headers all be 2025? Since these are new files and should reflect their current status.

Pri-yan-shu:
@sletz Fair enough. I kept the `pw_thread_loop *loop` for `pw_thread_loop_destroy(loop)` in the destructor, but all pipewire specific code has been written by looking at pipewire's example programs, specificly [https://gitlab.freedesktop.org/pipewire/pipewire/-/blob/master/src/examples/midi-src.c](url) and [https://gitlab.freedesktop.org/pipewire/pipewire/-/blob/master/src/examples/audio-dsp-filter.c](url).

Pri-yan-shu:
@dromer I began with jack architecture files as base and didn't knew much about writing licence info so I kept the dates as is assuming maintainers would take care of that. I would update the dates of the files.

Pri-yan-shu:
@sletz I have cleaned up the code as much as I could, and fixed the licence dates. I replaced unnecessary state with requesting variables on the fly. I need `pw_thread_loop *loop` for locking and destruction (no means to get `pw_thread_loop` from `pw_filter`). Pipewire doesn't take the ownership of `filter_events` and `filter_listener` so I need to keep them as class members as well. I have tested midi and polyphonic support working fine. If you find any issues with the code point them out, I'll fix them.

sletz:
Thanks. 

- since I'm not a PW user at all, I would love you to properly comment the PW related code (threading, filter...) everywhere needed
-  ans also could you format the code with our naming convention, like using `fLoop` instead of `loop` (fields like `fXXX`...)

Pri-yan-shu:
Done

sletz:
Thanks. 

Can you check why this happens  ?: 

```
faust2pwconsole organ.dsp
/usr/local/bin/faust2pwconsole: line 111: [: =: unary operator expected

```

Pri-yan-shu:
The line is `if [ $POLY = "POLY2" ]; then`. If POLY is not assigned any value beforehand the line evaluates to `if [ = "POLY2" ]`, hence the error. I was getting same error on upstream faust2jackconsole so I assumed it to be fine. I have fixed it now, assigning `POLY="POLY"` at beginning like other scripts.

sletz:
Thanks a lot. 

Merged and Faust version raised to 2.80.3.

PR #1132, opened 2025-04-03
Fix/chaos stratus installer

bassmanitram:
The installer scripts inadvertently used some bash extensions to posix shells. This made the scripts invalid when running with posix shells that are not bash. Unfortunately Ubuntu has, for some weird reason, decided that the default shell - `/bin/sh` shall be `dash` not `bash`. And WSL is essentially Ubuntu! Which means the scripts don't work on Windows.

This PR replaces Bash specifics with their Posix (and Dash-compatible) alternatives.

It is fairly urgent to get this out to the Faust IDE because the main use case of this is building an installer from the IDE.


bassmanitram:
Apparently rebasing never works for me - closing

PR #1138, opened 2025-04-20
rust: move dsp struct

crop2000:
move dsp struct to the beginning of the exported code so that the use can use the architecture file to pass derive macros


sletz:
@bluenote10 OK for you ?

bluenote10:
> @bluenote10 OK for you ?

Yep, that should be fine.

bluenote10 (approved):

PR #1141, opened 2025-04-30
Use std::unique_ptr to manage the lifetime of midi_handler in DspFaust

kisielk:


sletz:
Thanks, but we probably have to check that this change in `MidiUI` does not break other use-cases.

PR #1143, opened 2025-05-05
fix  #737

### LINKED ISSUE #737: Recursively defined tables generating incorrect code
```
import("stdfaust.lib");
process = rdtable(1000, os.osc(500), 10);
```

produces incorrect code in all FIR backends.

or 

```
import("stdfaust.lib");
process = rdtable(1000,  rdtable(1000, os.osc(500), 10), 10);
```

produces incorrect code in -ocpp and all FIR backends.

crop2000:


crop2000:
this could be a suitable impulse test:

```
import("stdfaust.lib");
size = 100;
process = rdtable(size,os.osc(1/size),ba.time%size);
```

if you think this is not a good idea it would be better to have an error then broken resulting code.

sletz:
But the output is still all zero ?

crop2000:
the fix still does not work for -it

crop2000:
blocked recursion for -it

crop2000:
i fixed the problem with the -it and the test case

crop2000:
the problem is that sample_rate is not correctly initialized

```
   pub fn instance_initmydspSIG0(&mut self, sample_rate: i32) {
        self.fConst0 = 441.0 / F64::min(1.92e+05, F64::max(1.0, (self.fSampleRate) as F64));
```

crop2000:
need to much fixing after #1152 and i have not enough use for it.

PR #1147, opened 2025-05-18
rust: fix unecessary mutablity for waveforms

crop2000:
in rust: currently waveforms are stored in a mutable static variable that doesn't need to be mutable. this pr fixes this.


sletz:
Thanks.

Issue #1005, opened 2024-03-24
Rust backend: mutable reference to mutable static is deprecated


bluenote10:
Starting with Rust version 1.77 (released a few days ago), there is a new compiler warning related to "mutable reference to mutable static". 

Given a simple Faust DSP like

```faust
import("stdfaust.lib");

envelope(gate) = gate : en.adsr(0.01, 0.01, 0.3, 0.1);
osci(f) = (os.osc(f) * 0.5 + os.osc(f*2) * 0.25 + os.osc(f*3) * 0.125);
synth(gate, freq) = envelope(gate) * osci(freq);

process = synth;
```

The code generation will currently produce a pattern like this (omitting everything but the relevant stuff here):

```rust
static mut ftbl0GenOscillatorSIG0: [F32; 65536] = [0.0; 65536];

impl FaustDsp for GenOscillator {

    // ...

    fn class_init(sample_rate: i32) {
        let mut sig0: GenOscillatorSIG0 = newGenOscillatorSIG0();
        sig0.instance_initGenOscillatorSIG0(sample_rate);
        sig0.fillGenOscillatorSIG0(65536, unsafe { &mut ftbl0GenOscillatorSIG0 }); // <= problem here
    }

}
```

This produces the following warning:

```
warning: creating a mutable reference to mutable static is discouraged
   --> faust/src/gen_oscillator.rs:156:52
    |
156 |         sig0.fillGenOscillatorSIG0(65536, unsafe { &mut ftbl0GenOscillatorSIG0 });
    |                                                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^ mutable reference to mutable static
    |
    = note: for more information, see issue #114447 <https://github.com/rust-lang/rust/issues/114447>
    = note: this will be a hard error in the 2024 edition
    = note: this mutable reference has lifetime `'static`, but if the static gets accessed (read or written) by any other means, or any other reference is created, then any further use of this mutable reference is Undefined Behavior
    = note: `#[warn(static_mut_refs)]` on by default
help: use `addr_of_mut!` instead to create a raw pointer
    |
156 |         sig0.fillGenOscillatorSIG0(65536, unsafe { addr_of_mut!(ftbl0GenOscillatorSIG0) });
    |                                                    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

Note however, that this is only a warning _for now_. In the future it will become a hard error, so just silencing the warning is not an option. A bit more background can be found here:

[...]

PR #1150, opened 2025-05-31
rust: changes to allow 2024 edition

crop2000:


crop2000:
@bluenote10  @sletz this is the same pr as the one mentioned in #1005 it changes the scope of what is considered unsafe rust code and with this solves the warning that turned into an error in the 2024 edition of rust. this change is backwards compatible to older rust editions.

bluenote10:
I would need to take a closer, but on first glance I'm somewhat torn: As far as I can see the error in the 2024 edition is a genuine one, i.e., the current logic has undefined behavior. Just silencing it with more `unsafe` is technically unsound, because one should only use `unsafe` if one has _other means_ of guaranteeing that the code is valid, which is (likely?) not the case here. It could make it harder to actually come up with a proper solution, and could even hide other UB issues?

crop2000:
The code is valid if you can garanty that you don't initialize the class while you use it elsewhere. To get close to a sound solution one needs to deprecate the init method.

bluenote10:
> The code is valid if you can garanty that you don't initialize the class while you use it elsewhere.

The "if" in that sentence is exactly the reason why it is unsound. For instance, from the [Rustonomicon](https://doc.rust-lang.org/nomicon/working-with-unsafe.html):

> We say that such a correct unsafely implemented function is _sound_, meaning that safe code cannot cause Undefined Behavior through it [...].

Similarly, if we'd stick to the [safety comments convention](https://std-dev-guide.rust-lang.org/policy/safety-comments.html), we'd exactly lack a justification why that unsafe block is safe as far as I can see. As long as there is a condition like _if you can garanty that you don't initialize the class while you use it elsewhere_ then you would have to pass the `unsafe` to the outside.

I'd assume that would be pretty inconvenient, but the warning/error is genuine, and just claiming safety where there isn't doesn't really solve the problem. I need to find some time to look into the other ideas mentioned in https://github.com/grame-cncm/faust/issues/1005.

crop2000:
The current implementation (with and without my changes) has the following relevant qualities:
Tables 
1. have no overhead when accessed for reading (no atomic locks etc.)
2. can be read by multiple instances
3. can be read without overhead from multiple threads
4. can be initialized using regular functions.
Right now I have no idea how to archive this without major changes to Faust without mutable static.

If we accept that we have a mutable static we should restrict the possibility to mutate it. We could add an additional static variable that counts the instances of dsp objects and blocks mutation of the variable via `class_init` if instances exists. This change would be a breaking change.

Currently the re-initialization of class variables has no reasonable effect. Because sample rate changes do not affect tables: #1151 . so one could simple not execute the content of class_init() if it had been already initialized (also using a static atomic variable). this solution is backward compatible. But it would imply that #1151  stay unsolved or are solved in a way that limits the use of Sample rate. 

(i think issue #737 is also related because the use of sample rate becomes more likely if one has more complex table filling functions.)

crop2000:
if we wrap those ideas into a type and provide functions then the meaning doesn't really change just the scope of the unsafe clauses. 
here realized for solution B
```
static ftbl0SinOscSIG0_th: UnsafeTableHolder = UnsafeTableHolder {
    table: UnsafeCell::new([0.0; 65536]),
    lock: AtomicUsize::new(0),
};
fn ftbl0SinOscSIG0_th_ref() -> &'static [f32] {
    unsafe { &mut *ftbl0SinOscSIG0_th.table.get() }
}
fn ftbl0SinOscSIG0_th_mut_ref() -> &'static mut [f32] {
    assert!(
        !(ftbl0SinOscSIG0_th.lock.load(Ordering::Relaxed) > 0),
        "don't try"
    );
    unsafe { &mut *ftbl0SinOscSIG0_th.table.get() }
}
```

bluenote10:
By using `thread_local` it would be possible to avoid any `unsafe` altogether. Here is a small example. Given the following dsp code:

```faust
import("stdfaust.lib");

envelope(gate) = gate : en.adsr(0.01, 0.01, 0.3, 0.1);
synth(gate, freq) = envelope(gate) * os.osc(freq);

process = synth;
```

I've made the following manual edits to the generated code ([full code example](https://github.com/bluenote10/RustFaustExperiments/blob/master/rust/src/bin/osci_modified.rs)):

Instead of using plain/global statics, I'm using a thread-local static:

image

The `fillDspSIG0` method doesn't have to change at all, it can still take a `table: &mut [FaustFloat]` as argument, filling it as before. The usage in `class_init` becomes:

image

The usage inside `compute` comes down to `ftbl0DspSIG0.with_borrow(|ftbl0DspSIG0_borrow| { ftbl0DspSIG0_borrow[...])`. For performance reasons it would be good to do the borrow not on every sample in the for loop, but rather pull the borrow out of the loop like this:

image

Notice that there are no `unsafe` usages needed at all, because everything can be done entirely in safe Rust.

The only assumption is that the DSP is consistently used from the same thread. Considering that this is pretty much the standard in audio programming, I'd be tempted to go for a fully safe thread local solution instead of trying to cook up a sound unsafe usage.

crop2000:
Are interested to implement this?

crop2000:
I see that you use RefCell adds the overhead of a atomic check at every access that's not good. and not necessary.

bluenote10:
> I suggested checks in this discussion.

Yes, and they looked promising, but they essentially looked like an attempt at re-implementing what e.g. `Mutex` already provides in the standard library. Getting these things right is hard. The general recommendation is to not even try, but stick to the safe high-level abstractions the standard library provides. Since this use case looks pretty much standard, there probably no good reason to mess with `unsafe` directly.

compiler/generator/rust/rust_code_container.cpp
```diff
@@ -134,22 +134,24 @@ void RustCodeContainer::produceInternal()
...
-        *fOut << "pub fn fill" << fKlassName
-              << subst("(&mut self, $0: i32, table: &mut[i32]) {", counter);
+        *fOut << "fn fill" << fKlassName
+              << subst("(&mut self, $0: i32, mut table: *mut[i32]) { unsafe{", counter);
     } else {
         tab(n + 1, *fOut);
-        *fOut << "pub fn fill" << fKlassName
-              << subst("(&mut self, $0: i32, table: &mut[FaustFloat]) {", counter);
+        *fOut << "fn fill" << fKlassName
+              << subst("(&mut self, $0: i32, mut table: *mut[FaustFloat]) { unsafe{", counter);
```
bluenote10, line 146: Is the `mut` before `table` actually needed?

At least my example was compiling without it, and as far as I can see the pointer itself isn't mutated, right?

compiler/generator/rust/rust_code_container.cpp
```diff
@@ -273,7 +275,7 @@ void RustCodeContainer::produceClass()
         tab(n + 1, *fOut);
         *fOut << "#[cfg_attr(not(target_os = \"windows\"), link(name = \"m\"))]";
         tab(n + 1, *fOut);
-        *fOut << "extern \"C\" {";
+        *fOut << "unsafe extern \"C\" {";
```
bluenote10, line 278: What is the reason for adding these `unsafe` here?

bluenote10 (commented):
> This pr does not make the code less correct than before. Are you open that this change (without the proposed locks) is merged until you implemented your solution?

If it simplifies things for you, we can go ahead and merge it.

Conceptually it's just a step in the wrong direction. You're right that the code was unsound all the time. The purpose of this warning/error is however to point out that it is unsound, and to solve it properly. Adding an unsound `unsafe { ... }` block like we do now is basically what the warning/error was trying to prevent. In a sense, an invalid logic with a warning is marginally "better" than an invalid logic where the warning has been incorrectly suppressed.

compiler/generator/rust/rust_code_container.cpp
crop2000: Edition 2024 changed rule to make it explicit.

crop2000:
the only change leftover is this the change caused by this: https://doc.rust-lang.org/nightly/edition-guide/rust-2024/unsafe-extern.html 
its backwards compatible

bluenote10 (approved):

bluenote10:
@crop2000 Was about to mention this, thanks for taking care of it!

sletz:
Done here: https://github.com/grame-cncm/faust/commit/a7b8173887cb1f9c30c2addc560b806d4638cd96

PR #1174, opened 2025-10-03
JSFX Poly initialization 

johannphilippe:
JSFX backend : Added code to initialize polyphonic parameters (freq, vel, gain, gate etc) for each voice.


sletz:
The JASFX backend is somewhat broken since this commit: backend JSFX à un problème depuis ce commit: https://github.com/grame-cncm/faust/commit/70bcbc6415bacbbd3572a610938248d466a9d677.

See: https://gist.github.com/sletz/dc9aac55cea51356b13892e1a97ba8d4

johannphilippe:
I'm not able to reproduce this failing test locally. Almost everything passes like before (except bs, grain3, priority and math) 
Though, it is also weird that it wasn't issuing before this commit (since parameters were not initialized)

I tested manually one of the issuing here (karplus) : it behave as expected on Reaper JSFX.

PR #1179, opened 2025-11-13
modify gGlobal, add per-instance garbage collection

DBraun:
I was trying to use nanobind instead of pybind11 with DawDreamer. Claude identified`gGlobal` as a potential reason that Python was exiting with a 134 error code once a program had finished. This PR is a potential fix. I haven't done the full loop of creating libfaust, building DawDreamer etc, but I thought I'd just share it first.


DBraun:
Good news, this wasn't the issue in DawDreamer.

PR #1181, opened 2025-11-21
faust2esp32 update 5.x

thopman:
- update faust2esp32 for 5.X. 
- removed midi dependency on jdks-midi. midi works without any external library
- updated ac101 driver for esp-idf 5.x


sletz:
Merged, thanks.

PR #1187, opened 2025-11-22
Extend support for ESP32 targets

pera:
This PR extends the current support for other ESP32 targets (e.g. ESP32-S3), allows for custom I2S pin configurations and also simplifies integration with DACs that don't require a codec driver (e.g. PCM5100).

### DIFF
--- a/architecture/esp32/esp32.cpp
+++ b/architecture/esp32/esp32.cpp
@@ -89,7 +89,7 @@ list<GUI*> GUI::fGuiList;
 ztimedmap GUI::gTimedZoneMap;
 #endif
 
-AudioFaust::AudioFaust(int sample_rate, int buffer_size)
+AudioFaust::AudioFaust(int sample_rate, int buffer_size, i2s_pin_config_t pin_config)
 {
 #ifdef NVOICES
     int nvoices = NVOICES;
@@ -101,7 +101,7 @@ AudioFaust::AudioFaust(int sample_rate, int buffer_size)
     fUI = new MapUI();
     fDSP->buildUserInterface(fUI);
     
-    fAudio = new esp32audio(sample_rate, buffer_size);
+    fAudio = new esp32audio(sample_rate, buffer_size, pin_config);
     fAudio->init("esp32", fDSP);
     
 #ifdef SOUNDFILE
--- a/architecture/esp32/esp32.h
+++ b/architecture/esp32/esp32.h
@@ -42,6 +42,14 @@ class esp32_midi;
 class SoundUI;
 #endif
 
+#define DEFAULT_I2S_PIN_CONFIG { \
+    .mck_io_num = I2S_PIN_NO_CHANGE, \
+    .bck_io_num = I2S_PIN_NO_CHANGE, \
+    .ws_io_num = I2S_PIN_NO_CHANGE, \
+    .data_out_num = I2S_PIN_NO_CHANGE, \
+    .data_in_num = I2S_PIN_NO_CHANGE \
+}
+
 class AudioFaust
 {
     private:
@@ -59,7 +67,7 @@ class AudioFaust
 
     public:
     
-        AudioFaust(int sample_rate, int buffer_size);
+        AudioFaust(int sample_rate, int buffer_size, i2s_pin_config_t pin_config = DEFAULT_I2S_PIN_CONFIG);
         ~AudioFaust();
     
         bool start();
--- a/architecture/faust/audio/esp32-dsp.h
+++ b/architecture/faust/audio/esp32-dsp.h
@@ -143,7 +143,7 @@ class esp32audio : public audio {
     
     public:
     
-        esp32audio(int srate, int bsize):
+        esp32audio(int srate, int bsize, i2s_pin_config_t pin_config):
         fSampleRate(srate),
         fBufferSize(bsize),
         fNumInputs(0),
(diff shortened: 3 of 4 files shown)


sletz:
Thanks @pera. @thopman can you have a look ?

thopman:
Thanks for working on this @pera! I appreciate the thought behind making the code more flexible.
I have some concerns about this approach though:

Logic issue: The pin_config parameter is being overwritten inside the constructor by the #if blocks, so custom configurations passed in won't actually be used when board defines are set.

Missing default: The original #else case with working default pins (bck: 33, ws: 25, data_out: 26, data_in: 27) was removed. Now DEFAULT_I2S_PIN_CONFIG is all I2S_PIN_NO_CHANGE, which won't work for boards without a specific define.

Design concern: I'm not sure this flexibility is needed in practice. Pin configuration is really a compile-time concern - you set it once for your hardware and it never changes. You're not swapping codecs at runtime. The current pattern with board defines makes it clear what pins are being used and works well for the use case.
If someone has a custom board, they're already going to be editing the code to add their codec initialization. Adding a #elif for their board or tweaking the default is straightforward and keeps everything explicit.
The added runtime parameter creates ambiguity about where pins are actually configured and adds complexity for a scenario that doesn't really come up.
What do you think? I'm inclined to stick with the current implementation, but I'm open to discussing if there's a use case I'm missing.

pera:
Hi @thopman thanks very much for reviewing and your comments. Given that there are a few things happening in this PR I will list them here:

1.  Put `FUNC_GPIO0_CLK_OUT1` (_soc/esp32/register/soc/io_mux_reg.h_) behind a flag: Without this change `faust2esp32` output can only be compiled for `ESP32` target chip (the name is confusing so just to be clear, this is not referring to the family but the original ESP32). For example, my target `ESP32S3` doesn't compile because this define doesn't exist in _[soc/esp32s3/register/soc/io_mux_reg.h](https://github.com/espressif/esp-idf/blob/130fdc7ce755207811740f1163e6f394e0e86c0b/components/soc/esp32s3/register/soc/io_mux_reg.h)_. Maybe this could be configurable but I don't believe it's strictly necessary.
2. Regarding the possibility of pin configuration being overwritten, this is a fair concern and I do agree that maybe the proposed change is not the best option but I do believe this should be configurable and also that the current behaviour is confusing.
3. I think part of the issue is that `faust2esp32` defaults to `WM8978` / `TTGO_TAUDIO` hardcoding its pinout parameters. As a new user this was a bit confusing for me (my guess is that the faust2esp32 started with this platform as the only target and then added a few other ones but wanted to keep backward compatibility, is this correct?).

Maybe a better option for point 2 could be to make GPIO pins configurable through Kconfig and sdkconfig. What do you think?

My intention with this PR is to make it as easy as possible for people wanting to use an esp32-s3 together with pcm5100 or pcm5102 breakout boards, DACs that don't require codec initialisation code. I think this combination is very common because you can buy them for about £10 on many shops.

pera:
Thanks very much for your thoughts @thopman, I have created a new PR for the `FUNC_GPIO0_CLK_OUT1`: https://github.com/grame-cncm/faust/pull/1188

Regarding the PIN defaults:

- The reason I suggested Kconfig + sdkconfig.defaults was because I wrongly believed Faust already used it. A similar alternative is having those values as macro definitions and then allow users to use CMake's `add_compile_definitions`, for example. What do you think?
- I have the feeling we are using Faust in different ways: I installed Faust through my Linux distribution official package manager repository. Faust is officially available in Arch Linux, Debian, Ubuntu and provably other distros. I am not aware of a nice way to configure GPIO PINs for such install (i.e. without having to either modify system wide Faust files or having to manually patch output code). I imagine maybe you are working on a fork and making changes for your specific targets, so it's not a big deal for you to hardcode a few values, but I think this can create significant friction for users who installed Faust through their Linux distro package manager, as it force them to be familiar with Faust's codebase and also having to maintain these changes each time the package is updated.
- If you still prefer to use hardcoded PIN configuration the only thing I would ask if I may is to use lower GPIO numbers given that ESP32-S3 is common in very small form factors (e.g. Seeed Studio XIAOs have only 14 PINs).

sauloverissimo:
Hi everyone, I've been working on a project using Faust with a LilyGO T-Display-S3 and a PCM5102A DAC and ran into the same compilation issue described here. I already use this kind of pin mapping in other projects with this board, like this example I put together: https://github.com/Xinyuan-LilyGO/T-Display-S3/tree/main/examples/T-Display-S3-Music-Explorer

While getting it to work I ended up extracting the pin configuration into a separate header (`esp32-board-pins.h`) that maps known boards to their I2S pins. The existing boards (TTGO_TAUDIO, A1S_BOARD, LYRA_T) are preserved exactly as they are, backward compatibility is maintained in the default path, and the `#ifndef` guards in the else branch let users on stock distro installs override individual pins via `-DI2S_BCK=x` in CMake without touching any installed file.

The change to `esp32-dsp.h` is minimal — the four `#if/#elif` blocks collapse into a single `pin_config = { I2S_MCK, I2S_BCK, I2S_WS, I2S_DOUT, I2S_DIN }`. I also included the `CONFIG_IDF_TARGET_ESP32` guard for `FUNC_GPIO0_CLK_OUT1` from @pera's PR #1188, since it's needed for S3 compilation anyway.

The code is working and tested here. I'm available to open a PR if you think this direction makes sense — wanted to check first before going ahead.

PR #1192, opened 2025-12-16
move derive of default-boxed to architecture file

crop2000:
Since #1138 it is possible to put derive macros right before the included code and they will be applied to the dsp struct because of this change the default-boxed can be moved out of the code generation.


sletz:
@bluenote10 OK for you ?

sletz:
Merged, thanks.

Issue #1118, opened 2025-02-09
Add flag for disabling generation of `-lm` ffi code when compiling for `-lang rust`


mitchmindtree:
I could be wrong, but it seems there's no easy way to do this at the moment. Please correct me if I'm wrong!

This would be useful when compiling faust dsp code to rust that is to be compiled to the `wasm32-unknown-unknown` target.

sletz:
So you mean converting Rust generated code directly to wasm ? Why not doing that when compiling the Rust   code itself ?

mitchmindtree:
When compiling `.dsp` to `.rs` using `faust -lang rust foo.dsp`, `faust` generates a `ffi` module with functions that expect to link to lib `m`, a system math library.

Here's an example of this output when compiling the sine.dsp example to rust:

```
// ...

mod ffi {
        use std::os::raw::c_float;
        // Conditionally compile the link attribute only on non-Windows platforms
        #[cfg_attr(not(target_os = "windows"), link(name = "m"))]
        extern "C" {
                pub fn remainderf(from: c_float, to: c_float) -> c_float;
                pub fn rintf(val: c_float) -> c_float;
        }
}

// ...
```

When compiling this rust code generated by `faust` to `wasm32-unknown-unknown`, linking fails as we can't link to this system library for a wasm build.

Thus, if you're using `faust` to generate Rust code that you want to compile to wasm, you need to manually delete this section from the generated output and insert code that uses [the WASM-compatible `libm` crate](https://crates.io/crates/libm) instead.

It would be handy if there were a flag to avoid linking to lib `m` to avoid the need for a hack to remove it.

sletz:
Ah OK ! Done in https://github.com/grame-cncm/faust/commit/ad2e1566fa9fd3423e8a1ee7773f701b512c6a79 and version raised  to 2.79.0 https://github.com/grame-cncm/faust/commit/d46a01a6059928a91115996c8468bcd7e8202dce.

sletz:
@mitchmindtree like using Faust in [nanou](https://github.com/nannou-org/nannou) ?

crop2000:
@mitchmindtree  shouldn't it be possible to solve this issue with compile flags in rust similar to the treatment of the "windows" target?

bluenote10:
Getting rid of the libm dependency was also something I was looking for.

[...]

PR #1196, opened 2025-12-28
feat: improve rust codegen for wasm target

bluenote10:
This PR improves the Rust codegen for targeting wasm, following up on ideas from #1118.

Basically now the generated code works for both non-wasm and wasm compilation. (My requirement was to have the _same_ generated code work under both.)

To demonstrate, the generated code now looks like this:

```rust
#[cfg(not(target_arch = "wasm32"))] // Compile ffi bindings only on non-wasm targets
mod ffi {
    use std::os::raw::c_float;
    // Conditionally compile the link attribute only on non-Windows platforms
    #[cfg_attr(not(target_os = "windows"), link(name = "m"))]
    unsafe extern "C" {
        pub fn remainderf(from: c_float, to: c_float) -> c_float;
        pub fn rintf(val: c_float) -> c_float;
    }
}
fn remainder_f32(from: f32, to: f32) -> f32 {
    #[cfg(not(target_arch = "wasm32"))] // non-wasm targets use ffi bindings
    unsafe { ffi::remainderf(from, to) }
    #[cfg(target_arch = "wasm32")] // wasm relies on libm
    libm::remainderf(from, to)
}
fn rint_f32(val: f32) -> f32 {
    #[cfg(not(target_arch = "wasm32"))] // non-wasm targets use ffi bindings
    unsafe { ffi::rintf(val) }
    #[cfg(target_arch = "wasm32")] // wasm relies on libm
    libm::rintf(val)
}
```

This follows @mitchmindtree's idea in https://github.com/grame-cncm/faust/issues/1118#issuecomment-2816626272 and works for both use cases:

- native targets: They use the `ffi` and linking to `lm` as before.
- wasm target: They delegate to [`libm`](https://docs.rs/libm/latest/libm/index.html), a pure Rust crate offering the same functionality. Of course, wasm users need to add `libm` as a dependency.

---

Note that this makes the `-rnlm`/`--rust-no-libm` command line args (and `gRustNoLibm` internally) more or less obsolete. Removing it would be a breaking change though for someone expecting these command line arguments to exist. I didn't remove it yet, but if the breaking change is considered minor enough, we could do so in a follow-up. This would be more or less the inverse of https://github.com/grame-cncm/faust/commit/ad2e1566fa9fd3423e8a1ee7773f701b512c6a79.

### DIFF
--- a/compiler/generator/rust/rust_code_container.cpp
+++ b/compiler/generator/rust/rust_code_container.cpp
@@ -295,6 +295,8 @@ void RustCodeContainer::produceClass()
     // Missing math functions
     // See: https://users.rust-lang.org/t/analog-of-c-std-remainder/59670
     if (gGlobal->gFloatSize == 1 && !gGlobal->gRustNoLibm) {
+        *fOut << "#[cfg(not(target_arch = \"wasm32\"))] // Compile ffi bindings only on non-wasm targets";
+        tab(n, *fOut);
         *fOut << "mod ffi {";
         tab(n + 1, *fOut);
         *fOut << "use std::os::raw::c_float;";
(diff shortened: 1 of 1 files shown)


sletz:
@crop2000 You added `-rnlm/--rust-no-libm` right ? So can we remove them ?

bluenote10:
> @crop2000 You added -rnlm/--rust-no-libm right ?

@sletz If I remember correctly @crop2000 had added the `-rnt`/`--rust-no-faustdsp-trait` (which is independent), but I believe `-rnlm`/`--rust-no-libm` was added by you in https://github.com/grame-cncm/faust/commit/ad2e1566fa9fd3423e8a1ee7773f701b512c6a79 more or less as a quick fix for https://github.com/grame-cncm/faust/issues/1118.

sletz:
OK, so better wait for @mitchmindtree feedback then ?

bluenote10:
> OK, so better wait for @mitchmindtree feedback then ?

My idea was to merge this approach and release it so that it's easy for people to use it and get rid of their `--rust-no-libm` usages first. Then we can still make the decision whether to remove it at a later point based on feedback.

sletz:
OK done, thanks.

Issue #1151, opened 2025-06-02
fSampleRate is not initialized in SubContainers


crop2000:
```
import("stdfaust.lib");
process = rdtable(100,ma.SR,ba.time%100);
```

produces:

```
class mydspSIG0 {
	
  private:
	
	int fSampleRate;
	float fConst0;
	
  public:
	
	int getNumInputsmydspSIG0() {
		return 0;
	}
	int getNumOutputsmydspSIG0() {
		return 1;
	}
	
	void instanceInitmydspSIG0(int sample_rate) {
		fConst0 = std::min<float>(1.92e+05f, std::max<float>(1.0f, float(fSampleRate)));
	}
	
	void fillmydspSIG0(int count, float* table) {
		for (int i1 = 0; i1 < count; i1 = i1 + 1) {
			table[i1] = fConst0;
		}
	}

};

static mydspSIG0* newmydspSIG0() { return (mydspSIG0*)new mydspSIG0(); }
static void deletemydspSIG0(mydspSIG0* dsp) { delete dsp; }

static float ftbl0mydspSIG0[100];
```

the subcontainer does not initialize the sample rate.

I would suggest to restrict the use of fSampleRate in tables. So that this problem doesn't stays unnoticed.

crop2000:
Because it would reflect what is currently possible and would allow simplification of the interface. I agree that my position is debatable and I will not further argue for it.

sletz:
So, no for "a flag that disables the use of sample rate in subcontainers"

sletz:
Fixes in 2.83.7 version.

PR #1201, opened 2026-01-18
Fix: Initialize fSampleRate in SubContainer instanceInit (Issue #1151)

git-lakshy:
when Faust generates code for SubContainers (helper classes like `mydspSIG0`), the `fSampleRate` member was declared but never initialized, causing undefined behavior as mentioned in #1151 .

## Cause
The main DSP class uses `generateSR()` (called via processFIR()) which correctly initializes `fSampleRate`:
However, **SubContainers** use `produceInternal()` which calls `generateInit()` directly **without calling** `generateSR()`, missing the initialization.

## Fixing the missing initialization
Centralized fix: Call `generateSR()` in `generateSubContainers()` before `produceInternal()`, ensuring all backends automatically initialize `fSampleRate` in SubContainer `instanceInit` methods
 
 ## Steps to reproduce
 ```
import("stdfaust.lib"); 
process = rdtable(100, ma.SR, ba.time % 100);
```
current output :

```
void instanceInitmydspSIG0(int sample_rate) {
	fConst0 = std::min<float>(1.92e+05f, std::max<float>(1.0f, static_cast<float>(fSampleRate)));
	}
```	
 c++ backend gives the similar desired output with this fix :

```
void instanceInitmydspSIG0(int sample_rate) {
    fSampleRate = sample_rate; // <-- now initialized 
    fConst0 = std::min<float>(1.92e+05f, std::max<float>(1.0f, static_cast<float>(fSampleRate)));
}
```

### DIFF
--- a/compiler/generator/c/c_code_container.cpp
+++ b/compiler/generator/c/c_code_container.cpp
@@ -185,6 +185,7 @@ void CCodeContainer::produceInternal()
           << "* dsp, int sample_rate) {";
     tab(n + 1, *fOut);
     fCodeProducer->Tab(n + 1);
+    generateSR();
     generateInit(fCodeProducer);
     generateResetUserInterface(fCodeProducer);
     generateClear(fCodeProducer);
--- a/compiler/generator/c/c_code_container.cpp
+++ b/compiler/generator/c/c_code_container.cpp
@@ -185,7 +185,6 @@ void CCodeContainer::produceInternal()
           << "* dsp, int sample_rate) {";
     tab(n + 1, *fOut);
     fCodeProducer->Tab(n + 1);
-    generateSR();
     generateInit(fCodeProducer);
     generateResetUserInterface(fCodeProducer);
     generateClear(fCodeProducer);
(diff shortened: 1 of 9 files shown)


sletz:
Thanks @git-lakshy for taking this issue. Several suggestions: 
- it is not really satisfactory to have to add `generateSR()` on all backends. As it is done for the main container in `processFIR()` , it would be better to have a single centralised point add `generateSR()`, for all containers. Can you rework your patch in this direction ?
- this kind of incorrect produced C++ code should be tested in the impulse-tests infrastructure, using tools like [Clang-Tidy](https://clang.llvm.org/extra/clang-tidy/index.html) and [AddressSanitizer](https://clang.llvm.org/docs/AddressSanitizer.html). This is not yet in place but would be a very valuable contribution in the  impulse-tests, and (in particular) would have detected this precise problem.  
- and `process = rdtable(100, ma.SR, ba.time % 100);` test case should be added in impulse-tests infrastructure, so that the expected code generated bug fix can be checked.

sletz:
Thanks a lot, I reworked and extended your PR in several commits to finally have 2.83.7 here https://github.com/grame-cncm/faust/pull/1201

git-lakshy:
glad to help!

Issue #1206, opened 2026-01-24
rust: allow overloading of float types in architecture file


crop2000:
It would be nice if the float types used by rust FaustFloat and F32 and F64 could be overloaded to use specialized floats.

# Usecases 

1) NoisyFloat, which panics on Nan for debugging.

2) a denormal resistant float.

3) a float that makes use of fast-math.

4) mix and match f32 and f64 where applicable. 

# Needed Changes

There are the following needed changes:

1) currently casts between F32/F64 and FaustFloat are disabled with the use of gFAUSTFLOAT2Internal = true, this needs to be disabled.

2) currently code-gen writes `type FaustFloat  = F32` this would have to be move to the architecture file.

3) only primitive types f32 and f64 can be casted using the `as` operator.. We need to introduce a Trait to provide a common casting interface that is implemented for the default cases f32,f64, i*, u*, and needs to be implemented for the custom types.

4) one needs to overload the casting visitor to make use of the casting trait.

5) one needs to fix initialization to make use of the casting trait.

6) one needs to make the implemented math functions more generic `arg: Into<f32>` to be casted from the custom type ... or move them into a FaustMath trait that provides rintf...  with provided implementation for f32/f64 

# Problems

1) because a cast from a literal to a custom float is realized via a trait method they cannot be applied on static variables (as used for tables) this could only be solved if all tables are wrapped into a LazyLock.

# Questions?

1) Is it ok to introduce more Traits that need to be provided via Architecture file? Or more general, is it ok to make the boilerplate code needed in the architecture file bigger?

2) The list of changes is quite long and i think most of the changes are only impactful if the final aim  (to be able to use custom floats) is reached. But one could nevertheless implement them step by step. for example a FaustMath trait could simplify the codegen. any opinions on this?

3) if the problem with tables is not solved one could still make use of the feature in cases where no tables are used, does this makes sense?

crop2000:
@bluenote10 
@obsoleszenz 
@magnetophon

sletz:
*Or more general, is it ok to make the boilerplate code needed in the architecture file bigger?* ==> I would say yes. In general, we prefer having simpler backends and moving the "complexity" in architecture files.

crop2000:
On my branch i implemented the first 2.5 points from above: 

1. casts inside the dsp between   F32/F64 and  FaustFloat are now enabled.

2. The line ` type FaustFloat  = F64` is removed from codegen. and is added to the architecture file.

[...]

PR #1208, opened 2026-01-25
Rust: allow FaustFloat and F32/F64 to be different.

crop2000:
First pr about the topic discussed in #1206.
It changes the Rust code gen to allow FaustFloat and F32/F64 to be a different primitive type. 
It makes the generated rust code behave more similar to the generated c++ code.


sletz:
That seems reasonable to me. @bluenote10 any drawbacks you could see ?

bluenote10:
If I understand it correctly that's a breaking change? I would need a bit more time to think about the general direction for #1206.

One thing that not clear to me yet (but I haven't had to time to think it through). Isn't the precision to be used defined by the command line arguments to faust? If so, shouldn't it be the responsibility of the codegen instead of the architecture file to pick the type?

crop2000:
in most languages it is the responsibility of the architecture file to define FAUSTFLOAT for example in c/c++.
beside that one has to  add `type FaustFloat = f32` or `type FaustFloat = f64` to the architecture file this is not changing the expected behavior.
(ubuntu testrun just failed because of a problem during the initialization of the github action)

sletz:
To clarify here, some historical context:
- in C/C++ we have the concept of internal format for the computation itself, which can be `-single `(the default) or `-double`, and the format of external samples (the audio buffers) and controllers, which is this `#define FAUSTFLOAT float (or double)` line that we have in the architecture file. The point was to be able to distinguish what the audio layer format supports  (when running using JACK/CoreAudio.etc), and what the internal DSP computation format is
- this distinction is somewhat complex to have in all backends, so some of them typically have the same internal/external format (LLVM IR, wasm.... etc..)
- this PR makes the Rust backend look like more like C/C++ ones, also because @crop2000 you have some additional [needs](https://github.com/grame-cncm/faust/issues/1206) that this PR would make easier to have yes ?

crop2000:
this pr provides primarily the same benefits as the current situation on c/c++. So that one can have a 64bit DSP operating on a 32bit buffer. I think this is practical.

crop2000:
an option to make this change backwards compatible is to keep the line but use a flag to disable its generation. but i would prefer the current solution to not increase complexity.

bluenote10:
Yes this sounds sensible. By the end of the week I could think about it more thoroughly and experiment how big the impact of the breaking change would be (likely manageable), if waiting a few more days would be fine?

EDIT: Oh, what about the failing impulse tests?

crop2000:
I will fix it tonight seems i forgot some change detail after rebase

bluenote10:
While I'm not really happy that the codegen spits out that many casts now, I think we can go for this.

> just as a reminder, in rust are noop casts like from f64 to f64 optimized away

While this _should_ be more or less true, it is also a bit over simplified, and rather tricky to verify.

I spend a few hours now trying to prove that the casts really don't have an effect. They definitely have an effect on the generated binary file itself, i.e., if you compile a DSP using the current `master-dev` branch vs this branch with default rustc flags you _will_ get a different binary content. After going down the rabbit hole of disassembly diffing and messing with various rustc flags I'd conclude that these changes _should_ only be on the symbol level (for whatever reason the casts mess with the "anonymous" symbol naming of llvm). That's at least for release builds, I didn't dare to investigate that for debug builds. Do you know if this can be relied upon in debug builds?

---

What I find a bit fishy: My expectation was that when using `-single` and `FaustFloat = F32` the generated code would be free of any `F64` occurrences. This was the case before, and would make sense to me, because the user explicitly set everything to "f32 only". I basically always omitted the `F64 = f64` type alias to _guarantee_ my dsps are f32 only. Unfortunately this doesn't seem to work any more, because for some reason the generated code does spit out a few `(... as F64) as FaustFloat` now, which require defining `F64` even in "f32 only" use case. Do we understand why that's the case? At first glance this seems to mainly come from slider usages.

sletz:
@crop2000 I don't this commit is necessary: https://github.com/grame-cncm/faust/pull/1208/changes/ca186a58c794f597ad6b4fdfc7a16d28576d8261  ?

crop2000:
@bluenote10 
did you see casts from ((X as F64) as FaustFloat) that that are not applied on literals?
For literals it is clear that these casts are done when consts are resolved.
The reason for the casting to double comes from Faust and becomes visible because of the commit that sletz mentioned. 
```
BlockInst
	StoreVarInst(Address(fHslider0, kStruct), CastInst("FAUSTFLOAT", Double(1e+03f)))
	StoreVarInst(Address(fHslider1, kStruct), CastInst("FAUSTFLOAT", Double(1.0f)))
EndBlockInst
```

crop2000:
i removed https://github.com/grame-cncm/faust/commit/ca186a58c794f597ad6b4fdfc7a16d28576d8261 now there are less casts of literals

sletz:
@bluenote10 can you check again  after the latest changes ?

bluenote10:
> @bluenote10 can you check again after the latest changes ?

Yep, I've re-verified: The generated assembly is 100% the same. So everything should be good, can be merged.

>  did you see casts from ((X as F64) as FaustFloat) that that are not applied on literals?

Yes, so far I only saw these `((X as F64) as FaustFloat)` on literals. Fortunately after reverting this particular commit, these unexpected `as F64` casts have disappeared. So I was able to remove the `F64` type definition again and work in "f32 only" as before :+1:

sletz:
Merge done.

PR #1209, opened 2026-01-26
lv2.cpp : Added helper function for proper escaping in generating ttl metadata

### LINKED ISSUE #1204: LV2 backend does not escape FAUST metadata strings when generating C++ and `.ttl` output
The LV2 backend writes FAUST metadata strings directly into generated C++ and Turtle files without escaping, 
due to which if metadata in a DSP file contains characters that need escaping at the C++ level (like backslashes) it leads to wrong metadata generation as those characters get dropped or altered before the LV2 metadata is generated. If we try to work around the issue by double-escaping (//), the LV2 backend still writes metadata into Turtle files without Turtle/RDF string escaping which can result in possible invalid `.ttl` output. 

for example compiling this Faust DSP with `faust2lv2`
```
declare name "Gain \\ Rain";
declare author "Julius Smith";
declare description "CryBaby effect from the Faust effect library";
declare version "1.0";

import("effect.lib"); 

drywet(fx) = _,_ <: *(d), *(d) , (fx : *(w), *(w)) :> _,_
with {
  d = (1-w);
  w = vslider("dry-wet[style:knob]", 0, 0, 1, 0.01);
};

wahwah = hgroup("", drywet(par(i,2, crybaby(wah))))
with {
  wah = vslider("wah[style:knob]",0.8,0,1,0.01) : automat(bps, 10, 0.0);
  // not sure what units these are...
  bps = vslider("tempo[style:knob]", 360, 0, 1920, 1);
};

process = wahwah;
```

produces the .ttl metadata file :
```
@prefix doap:  <http://usefulinc.com/ns/doap#> .
@prefix foaf:  <http://xmlns.com/foaf/0.1/> .
@prefix lv2:   <http://lv2plug.in/ns/lv2core#> .
@prefix ui:    <http://lv2plug.in/ns/extensions/ui#> .
@prefix epp:   <http://lv2plug.in/ns/ext/port-props#> .
@prefix atom:  <http://lv2plug.in/ns/ext/atom#> .
@prefix rdf:   <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
@prefix rdfs:  <http://www.w3.org/2000/01/rdf-schema#> .
@prefix units: <http://lv2plug.in/ns/extensions/units#> .
@prefix urid:  <http://lv2plug.in/ns/ext/urid#> .
<https://faustlv2.bitbucket.io/lv2>
       a lv2:Plugin ;
       doap:name "Gain \ Rain" ;  \\  <<----- resulting back slash 
       lv2:binary <lv2.so> ;
.
.     
.

```
This .ttl output file is not valid Turtle syntax
>Validating this .ttl file with rapper (Raptor RDF library) gives the output :
Turtle string error - illegal escape \  (#x20) in "Gain \ Rain""

git-lakshy:
# In this PR
Created a function `ttl_escape()` to ensures proper escaping of strings in generated `.ttl` metadata, used while writing metadata in `lv2_dyn_manifest_get_data()`
fixes #1204

### DIFF
--- a/architecture/lv2.cpp
+++ b/architecture/lv2.cpp
@@ -1801,6 +1801,21 @@ static unsigned steps(float min, float max, float step)
   return n;
 }
 
+static string ttl_escape(const string &s)
+{
+  string t;
+  for (char c : s) {
+    switch (c) {
+    case '\\': t += "\\\\"; break;
+    case '\t': t += "\\t"; break;
+    case '\n': t += "\\n"; break;
+    case '\r': t += "\\r"; break;
+    default:   t += c; break;
+    }
+  }
+  return t;
+}
+
 #if FAUST_META
 static bool is_xmlstring(const char *s)
 {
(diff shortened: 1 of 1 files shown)


sletz:
@agraef any comment here ?

sletz:
Thanks, reworked and merged in 2.83.10.

PR #1210, opened 2026-01-28
Fix RTAlloc of FaustClass in SuperCollider

capital-G:
Currently the server will crash if RTAlloc of FAUSTCLASS fails, since it is tried to place the FAUSTCLASS to the nullptr, which is undefined behavior.
This was discovered via https://github.com/supercollider/sc3-plugins/pull/416

While I am at it: I also fixed some memory leaks when the allocation failed.

edit: This needs some rework - do not merge yet!


sletz:
Thanks, is it ready to be merged?

capital-G:
> Thanks, is it ready to be merged?

Not yet - I'll ping you when its ready ;)

PR #1212, opened 2026-02-03
rust change import of c_float to use core (no-std)

crop2000:


compiler/generator/rust/rust_code_container.cpp
```diff
@@ -297,7 +297,7 @@ void RustCodeContainer::produceClass()
         tab(n, *fOut);
         *fOut << "mod ffi {";
         tab(n + 1, *fOut);
-        *fOut << "use std::os::raw::c_float;";
+        *fOut << "use core::ffi::c_float;";
```
bluenote10, line 300: Shouldn't line 352 be adapted as well for consistency (`use std::os::raw::c_double`)?

PR #1214, opened 2026-02-07
rust: change libm code generation so that using rust libm directly easier

crop2000:
instead of wrapping c-libm make use of the libm crate 

example of wrapping c-libm is kept in architecture_ecrnt.rs


crop2000:
@bluenote10 i guess you want to have a look

bluenote10:
I don't think we should do this. I'd assume not using the native functions and only supporting the pure rust version has worse performance (I think we checked this back in the day). Also, this just removes options for users, it's already perfectly possible to use the Rust `libm` with current codegen.

crop2000:
the problem emerged when i tried to use the generated code on a different target. 
one could use the old design in all architecture files as i did in architecture_ecrnt.rs
do you have code that shows the performance difference of c-libm vs rust-libm ?

crop2000:
@sletz @bluenote10 

The disagreement here is if we leave a generated block of code that does not depend on the actual dsp in the code gen or if we reduce the code gen and move more things to the architecture.

bluenote10:
> do you have code that shows the performance difference of c-libm vs rust-libm?

My benchmarks are over [here](https://github.com/bluenote10/RustFaustExperiments/), but currently none of them seems to use rint/remainder functionality, so they don't show it. @sletz What would be a minimal dsp example that would produce usages of rint/remainder that could be used as a benchmark?

> we reduce the code gen and move more things to the architecture

My main concern is that this leads to bad DX. If the codegen becomes more and more incomplete, the generated code is essentially broken by itself, throwing obscure compiler errors when users try to compile things. We simply don't have a clear interface definition between the architecture and the codegen. How should users know what they have to provide to have a 100% correct set up? The fact that usage of rint/remainder seems rare makes this even more awkward: Users could think that their architecture setup works, but then they change some dsp in a way that usages of rint/remainder are emitted in the codegen and suddenly they get cryptic compiler errors, and they are not trivial to resolve them without spotting this one architecture example buried somewhere in the faust repo.

We have two options:

- Keeping the codegen largely self-contained, i.e., it simply produces all the things it needs by itself.
- If we want to break the codegen apart, we need to figure out a mechanism that allows for a clear interface definition with easy to use defaults. I've been thinking about this, but I'm not sure if there's a trivial solution. There could be e.g. some kind of `FaustOps` trait that defines the functionality that is potentially used by the codegen. The codegen should emit this `FaustOps` trait interface definitions plus implementations of `FaustOps` for both `f32` and `f64` so that users don't have to re-invent the wheel for the standard float types. But then it gets awkward to provide different impls. There could be either command line switches for that, or the user needs to hook up the used traits somehow, or we go for a newtype patterns, i.e., the user would set `struct CustomFloat(f32); type FaustFloat = CustomFloat` and they could implement the trait manually for their `CustomFloat`. But these are all pretty big changes.

So far I don't have good ideas for the latter, which is why I prefer the simplicity of the self-contained codegen.

bluenote10:
An idea: We could have a command line switch that allows to switch from the _fully self-contained_ to a _bring your own impl_ codegen.

Regular users would get the fully self-contained codegen out-of-the box. The emitted stuff could serve as a basis and at the same time as interface documentation of what a user would have to provide manually when switching to the other mode.

When doing that, it would make sense to also emit the `F32` + `F64` + `FaustFloat` definitions in the "fully self-contained" mode to make it really fully self-contained (currently we already have a bit of a hybrid, because these definitions need to be injected already). The codegen could then emit some comments documenting the "codegen interface", i.e., at the top of the generated code in the fully self-contained mode we would emit:

```rust
// The following functions are provided as default implementations in the self-contained mode.
// When switching to the manual injection mode, it's the responsibility of the architecture file to
// provide semantically equivalent definitions.

// begin of self-contained injected block (codegen interface)

type F32 = f32;
type F64 = f64;
type FaustFloat = f32;

fn rint(...) ...
fn remainder(...) ...

// end of self-contained injected block (codegen interface)
```

crop2000:
i don't think it is beneficial to have such things in the code gen. I was motivated to open this pr after this comment https://github.com/grame-cncm/faust/issues/1206#issuecomment-3794848393 by @sletz .

bluenote10:
Not beneficial? The codegen is _broken_ otherwise and it is highly non-trivial for users to even be aware of that!

crop2000:
it is possible to document the code block that needs to be added to the architecture. it does not depend on the code gen at all.

crop2000:
@sletz i added a change to keep that stuff that bluenote10 likes to keep.

sletz:
@bluenote10 OK for you ?

crop2000:
@bluenote10  @sletz ?

bluenote10:
Sorry for the delay! I'll have time to look into it this weekend.

bluenote10:
Unfortunately I didn't get to do a thorough review (as in: fully verify the generated output, which unfortunately isn't straightforward to see from the git review diff), I've too much on my plate right now.

But let's merge, as far as I can see it should go in the direction I was alluding to.

I probably would have introduced a command line switch for the entire dependency section instead of something specific for libm (which is kind of an implementation detail), which could be further extended for #1206, but this could still be consolidated later I hope.

sletz:
Merged, version 2.85.1

PR #1232, opened 2026-03-01
Fix build errors with LLVM 22

khiner:
Three APIs were removed from `TargetOptions` in LLVM 22:
- `UnsafeFPMath`, replaced by flags
- `setFPDenormalMode()` method
- `createTargetMachine()` no longer accepts a string for the target triple - it now requires a `Triple`


sletz:
Thanks.

PR #1256, opened 2026-04-05
Remove std::less<CTree*> specialization

khiner:
I was hitting infinite recursion nondeterministically - turns out, these `CTree` pointers were inconsistent between comparisons somehow. I could hit this during a single compile sometimes, but on teardown/recompile I hit it always in my case. Can't really explain what was changing CTree values in a way that would change the `serial` of what it references during a single compile. Never really got to the bottom of it.

This was causing set ordering invariants to break, causing infinite recursion in `symlistVisit` - it uses a `set<Tree>` to detect cycles during signal tree traversal. With the corrupted set, already-visited nodes couldn't be found, so it cycled until the stack overflowed.

Using default pointer comparison fixes since it only compares addresses and avoids derefs that can return inconsistent values if what they point to changes. This is safe because hashconsing guarantees unique pointers per logical tree.

However, I think this would technically break compiler determinism, so maybe it's not the right fix, but it completely solved the crashes I was seeing - I could provide more context if needed.

### DIFF
--- a/compiler/tlib/tree.hh
+++ b/compiler/tlib/tree.hh
@@ -105,17 +105,6 @@ typedef CTree* Tree;
 
 typedef std::vector<Tree> tvec;
 
-namespace std {
-
-// The std::less <CTree*>comparison function is redefined to provide an unique and stable ordering
-// for all CTree instances and so maintain determinism.
-template <>
-struct less<CTree*> {
-    bool operator()(const CTree* lhs, const CTree* rhs) const;
-};
-
-}  // namespace std
-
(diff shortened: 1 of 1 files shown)


sletz:
Own but we cannot merge it since adding `std::less<CTree*>` was indeed added to fix and old an problematic "non-determinism" issue the the compilation chain. So I guess your issue is somewhere else.

khiner:
Gotcha, thanks - I'll close this PR then and look into the root cause more when I get a chance

PR #1259, opened 2026-05-21
architecture/rust: register N×M JACK ports based on DSP channel count

magnetophon:
Disclaimer: vibe-coded, but human tested.

The jack-float.rs and jack-double.rs JACK architecture files hard-coded 2-input / 2-output port handling: they always registered "in1", "in2", "out1", "out2" and called dsp.compute() with slices of length 2 regardless of what the generated DSP declared via get_num_inputs()/get_num_outputs(). Any DSP with a different channel layout (e.g. a 0-input, 4-output generator) panicked inside the realtime callback with

    panicked at src/main.rs:NNN: wrong number of output buffers
    thread caused non-unwinding panic. aborting.

This change makes both arch files:

  - read num_inputs/num_outputs from the DSP once, outside the realtime callback,
  - register Vec<Port<AudioInSpec>> and Vec<Port<AudioOutSpec>> sized accordingly (port names in1..inN, out1..outM),
  - inside the callback, build the &[&[f32]] / &mut [&mut [f32]] slice arrays dynamically and pass them to dsp.compute(),
  - derive the per-call frame count from outputs[0].len() (or inputs[0].len(), or ps.n_frames() for the 0-in 0-out generator edge case) rather than assuming input port 0 always exists.

For jack-double.rs, the f32↔f64 staging buffers are also made dynamic: Vec<Vec<f64>> sized to num_inputs/num_outputs, pre-allocated outside the realtime callback (previously they were allocated per process() call, which is unsafe in a JACK process callback).

Tested with several channel layouts (0×0, 0×4, 1×8, 2×2, 8×8) by compiling against the inherent-method signature emitted by the Rust code generator (rust_code_container.cpp: pub fn compute(&mut self, count: usize, inputs: &[impl AsRef<[FaustFloat]>], outputs: &mut [impl AsMut<[FaustFloat]>])).

Note: the FaustDsp trait declared at the top of these arch files specifies `fn compute(&mut self, count: i32, ...)` but is never implemented by the generated mydsp; the inherent method on mydsp is what actually gets called. The trait could either be removed or the codegen taught to implement it — left as a follow-up.

### DIFF
--- a/architecture/rust/jack-double.rs
+++ b/architecture/rust/jack-double.rs
@@ -104,7 +104,7 @@ pub trait UI<T> {
 <<includeIntrinsic>>
 <<includeclass>>
 
-const FRAMES_PER_BUFFER: u32 = 4096;
+const FRAMES_PER_BUFFER: usize = 4096;
 
 fn main() {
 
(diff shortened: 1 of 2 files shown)


sletz:
Thanks.

PR #1261, opened 2026-06-09
Embed the AssemblyScript backend in the wasm libfaust

petersalomonsen:
The AssemblyScript backend (`compiler/generator/assemblyscript/`, `-lang asc`) exists in the compiler but isn't embedded in the wasm `libfaust`: in `build/backends/regular.cmake`, `AS_BACKEND` is `COMPILER STATIC DYNAMIC` — it lacks the `WASM` keyword that `WASM_BACKEND` has. So a `libfaust-wasm` built via `make wasmlib` can't generate AssemblyScript; `generateAuxFiles(..., "-lang asc ...")` returns *"cannot find backend for 'asc'"*.

This adds `WASM` to `AS_BACKEND` so the backend is compiled into `libfaust-wasm`, matching `WASM_BACKEND`.

Verified in a clean CI build (no other patches): a `libfaust-wasm` built from this branch reports the asc backend present and generates valid `-lang asc` output for a range of real DSPs (DX7 algorithms etc.). One-line change.

Companion: grame-cncm/faustwasm#40 (a standalone ESM-loader fix needed when libfaust-wasm is rebuilt with a newer emscripten).


sletz:
Thanks, manually done in https://github.com/grame-cncm/faust/commit/7d95f0701de4349844c4ebec75f2a0dbc2b3a95d.

PR #1262, opened 2026-06-12
Changed types on inputs and outputs to compute! from overly constrain…

macd:
The types set on the inputs and outputs to compute! where set to the overly restrictive Matrix. This means that vectors that had been reshaped into a Matrix could not be used with compute!  By using AbstractMatrix, we fix this problem.


sletz:
Can you explain your use case ?

macd:
Yes, of course. I am directly managing the interaction with my external USB sound devices using some of the low level routines in libportaudio_jll and using ccall and I get the different channels interleaved in a single flat buffer. I reshape these buffers into the matrices that compute! expects. This change allows me to just drop in compute! into my scaffolding. Many thanks for the quick merge!

PR #1265, opened 2026-06-17
Fix faust2mathdoc to convert all mathdoc SVG diagrams to PDF

oneilltomhq:
## Summary

- `faust2mathdoc` only converted `process.svg` in each mathdoc SVG directory, but `--mathdoc` can emit named `<diagram>expr</diagram>` SVGs that the generated `.tex` references as PDFs.
- Switch the main loop to use the existing `convert_svgdirs2pdf` helper (same approach as `faust2mathviewer`) so every generated `.svg` is converted before `xelatex` runs.
- Add `tests/mathdoc/signalFlowPatternsMathdoc.dsp`, a small repro with multiple `<diagram>` tags.

## Problem

For DSP files with `<diagram>parallel_tank</diagram>` (and similar), `faust --mathdoc` writes SVGs like `parallel-*.svg` under `*-mdoc/svg/svg-*/`, and the TeX output references the corresponding PDFs. The script never created those PDFs, so `xelatex` failed with missing file errors:

```
! LaTeX Error: File `../svg/svg-01/parallel-0x...' not found.
```

## Repro

```bash
faust2mathdoc tests/mathdoc/signalFlowPatternsMathdoc.dsp
```

Before this fix, only `process.svg` was converted (1 PDF for 39 generated SVGs); the named diagram PDFs were missing and `xelatex` aborted. After the fix, all 39 generated SVGs are converted and the PDF builds successfully (46 pages).

## Test plan

- [x] `faust2mathdoc tests/mathdoc/signalFlowPatternsMathdoc.dsp` completes and writes `*-mdoc/pdf/*.pdf`
- [x] Verified 39 `.svg` files and 39 matching `.pdf` files under `svg/`
- [x] Existing `tests/mathdoc/rms.dsp` still builds (no regression)

## Notes

- Inkscape CLI flags left unchanged for compatibility.
- Missing `supertabular` (or similar LaTeX packages) on some systems is a separate dependency issue, not addressed here.

### DIFF
--- a/tests/mathdoc/signalFlowPatternsMathdoc.dsp
+++ b/tests/mathdoc/signalFlowPatternsMathdoc.dsp
@@ -1,0 +1,109 @@
+declare name "Signal Flow Patterns Mathdoc";
+declare version "0.1";
+declare author "oneilltomhq";
+declare license "BSD";
+declare description "Mathdoc tour of selected Faust example signal-flow patterns.";
+
+import("stdfaust.lib");
+
+<mdoc>
+\section{Signal-flow patterns in Faust examples}
+This document extracts small versions of three patterns used throughout the
+Faust examples:
+\begin{itemize}
+\item feedback delay networks and diffusers, as in \texttt{examples/reverb/freeverb.dsp}
+and \texttt{examples/reverb/fdnRev.dsp};
+\item crossfaded variable delay pitch shifting, as in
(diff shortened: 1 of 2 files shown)


sletz:
Is the PR ready now ?

oneilltomhq:
@sletz I believe so - maybe the test file is a little sloppy in its substance -  but the `faust2mathdoc` change allows multiple SVG diagrams, previously it was only allowing one

PR #1267, opened 2026-06-21
Fix division by zero in uiNumEntry when step is 0 (#1266)

### LINKED ISSUE #1266: Division by zero in QTUI.h in the uiNumEntry constructor when step==0.
I don't know the right solution to this, so I'm creating a new issue instead of a pull request. The division by zero happens  here:
https://github.com/grame-cncm/faust/blob/7d95f0701de4349844c4ebec75f2a0dbc2b3a95d/architecture/faust/gui/QTUI.h#L1138
if you have a line that e.g. looks like this:
`freq = nentry("freq",200,40,2000,0.01);`

vlerdf:
This Pull Request fixes a division by zero (and subsequent undefined behavior during the float-to-int cast) in the `uiNumEntry` constructor when `fStep` is set to `0.0`. 

It ensures that `decimals` is safely set to `0` if `fStep <= 0.0`.

Closes #1266


sletz:
@kmatheussen OK for you ?

kmatheussen:
Yes

PR #1268, opened 2026-07-03
faust2daisy - improvements and bug fixes

johannphilippe:
## Features 
* Soundfile primitive support: WAV files (PCM 16/24/32 bit, float 32 bit). Two modes:
  * default (`-qspi`): parsed at build time and inlined into QSPI flash, no runtime
    file I/O or dynamic allocation.
  * `-sd`: loaded from the SD card (`/soundfiles`) into SDRAM at startup. Files are
    uploaded with `faust2daisy -upload-sd` (temporary CDC uploader firmware).
  * Daisy Patch supported (specific mode, with Display and custom audio codec)
  * Daisy PatchSM supported 
 * Serial control through UART pins (both RX and TX are implemented), passing messages in the format "amp 0.72551" 
 * Hardware PWM output (defaults to 29Khz) wiht `pwm:D1`. 
 * PIN conflict checking on parser side (for example using `[adc:A0]` and `[gpio:D15]` on daisy seed won't work, since it would conflict, same pin)

## Fix 
* Midi input was not initialized 
* Python parser script DAC interpretation was wrong 
* MIDI CC was not working properly for checkboxes and buttons. Now it is working as Faust documentation says : returns 1 if CC is 127, returns 0 if CC is 0
* Python parser crashed on a DSP with UI controls when no configuration file was provided
* Fixed Patch screen bar ordering


johannphilippe:
Closed since there are git symbols pollution inside patch.json.

PR #1272, opened 2026-07-05
Fix the two C-API binary-operator bugs of #1264

### LINKED ISSUE #1264: libfaust 2.81.10 Two C-API binary-operator bugs
Hi, I'm working in a project with some ia that reported these two bugs. I think they are correct, for the first one it gave me a fix, it's at the end.

---

Two issues reachable from the **public C API**, both involving a binary-operator
primitive the box/signal layer mishandles. Each surfaces as a wrong/failed
compile or a hard `ASSERT` that aborts the host process.

Tested against libfaust 2.81.10 (LLVM backend). The fmod case was also reported
as present on master-dev.

---

## Bug 1 — `boxFmod()` builds `… : abs` instead of the 2-input fmod primitive

`CboxFmodAux(a, b)` (the applied `boxFmod()` primitive) produces a box that
compiles as if it were `(a, b) : abs`: a 1-input primitive fed 2 inputs, so
factory creation fails with an arity error. It looks like a copy-paste that
returns the `abs` primitive box instead of the fmod one.

Minimal reproduction (C Box API):

```c
createLibContext();
Box b = CboxFmodAux(CboxReal(5.0), CboxReal(3.0));   // expected: fmod(5,3) = 2
char err[4096];
llvm_dsp_factory* f = createCDSPFactoryFromBoxes("fmod", b, 0, NULL, "", err, -1);
destroyLibContext();
// f == NULL; err reports an arity/type mismatch consistent with `: abs`
```

Expected: a 2-input `fmod` box. Actual: behaves as `: abs`, factory creation
fails.

Workaround (for reference): build fmod from a source fragment
`CDSPToBoxes("process = fmod;")`, which yields the genuine 2-input primitive.

Faust (libfaust 2.81.10, still present on master-dev): boxFmod() in compiler/box_signal_api.cpp returns gGlobal->gAbsPrim->box() — apparently a copy-paste slip — so boxFmod(a, b) (and the C API's CboxFmodAux) builds (a, b) : abs, which then fails with  an arity error (1-input abs fed 2 signals) instead of yielding the 2-input fmod primitive; DSPToBoxes("process = fmod;") returns the correct primitive and works as a workaround.

---

## Bug 2 — logical right shift (`kLRsh`) aborts the signal type checker

`enum SOperator { kAdd, kSub, kMul, kDiv, kRem, kLsh, kARsh, kLRsh, … }` —
`kLRsh` is index 7. Building a signal that uses it (`CsigLRightShift`, i.e.
`sigBinOp` with `kLRsh`) and compiling it aborts: `sigtyperules.cpp` has no
case for that opcode. `kLsh` (5) and `kARsh` (6) type-check fine; only `kLRsh`
(7) is missing.

Minimal reproduction (C Signal API):

```c
createLibContext();
Signal a = CsigIntCast(CsigReal(3.0));
Signal b = CsigIntCast(CsigReal(1.0));
Signal out[] = { CsigLRightShift(a, b), NULL };
char err[4096];
createCDSPFactoryFromSignals("lrsh", out, 0, NULL, "", err, -1);   // aborts before returning
destroyLibContext();
```

Actual:

```
ASSERT : unrecognized opcode : 7
ASSERT : please report this message ... (file: sigtyperules.cpp, line: 154,
version: 2.81.10, options: -lang cpp -ct 1 -es 1 -mcd 16 -mdd 1024 -mdy 33
-single -ftz 0 )
```

Expected: type-infer the logical right shift like the other integer binary ops,
or reject it via the `error_msg` string instead of asserting and aborting the
host process.

smrg-lm:
Two fixes, one per bug of #1264:

**`boxCos()` / `boxFmod()`**: in `box_signal_api.cpp` both return `gGlobal->gAbsPrim->box()` instead of their own primitives, so `boxCos(x)` silently computes `abs(x)` and `boxFmod(b1, b2)` fails with an arity error. The signal-level `sigCos()`/`sigFmod()` are correct; only the box wrappers are affected.

**`kLRsh`**: the signal type checker (`arithmetic()` in `sigtyperules.cpp`) has no case for the logical-right-shift opcode, so any signal built with `sigLRightShift`/`CsigLRightShift` fails factory creation with `ASSERT : unrecognized opcode : 7` (and aborted the host process in 2.81.x). The added interval rule matches the arithmetic shift for non-negative operands and falls back to the full 32-bit range otherwise. Verified: `CsigLRightShift(int(3), int(1))` now compiles and renders 1.

Fixes #1264.

### DIFF
--- a/compiler/box_signal_api.cpp
+++ b/compiler/box_signal_api.cpp
@@ -1468,7 +1468,7 @@ LIBFAUST_API Tree boxLog10()
 }
 LIBFAUST_API Tree boxFmod()
 {
-    return gGlobal->gAbsPrim->box();
+    return gGlobal->gFmodPrim->box();
 }
 LIBFAUST_API Tree boxFloor()
 {
@@ -1484,7 +1484,7 @@ LIBFAUST_API Tree boxExp10()
 }
 LIBFAUST_API Tree boxCos()
 {
-    return gGlobal->gAbsPrim->box();
+    return gGlobal->gCosPrim->box();
 }
 LIBFAUST_API Tree boxCeil()
 {


sletz:
Thanks.

PR #1274, opened 2026-07-19
Add NNX backend and migrate JAX backend to Linen

DBraun:
> [!NOTE]
> Lots of AI was used in this PR, including an intense Fable 5 xhigh "workflow" review of the PR, but I have reviewed it all. I've been iterating on an NNX backend for several months, and this current draft is being used in a downstream "faustax" library, which I intend to release within a few months max.

**One line summary: Move the current JAX backend to `-lang linen` and add a new NNX backend with `-lang nnx`, along with architecture python files for each.**

JAX/Flax code generators share ~95% of their codegen:

- `-lang nnx` emits a Flax [NNX](https://flax.readthedocs.io/en/stable/) module (`nnx.Module`) with UI parameters as `nnx.Param` leaves (normalized to [0, 1]) and DSP state carried through an explicit dict, scanned per sample with `nnx.scan`.
- `-lang linen` (the previous `-lang jax` backend, renamed) emits a Flax [Linen](https://flax-linen.readthedocs.io/en/latest) module with parameters registered via `self.param` and a single carry dict scanned with `jax.lax.scan`.

Compiler: `compiler/generator/nnx` and `compiler/generator/linen` code containers and instruction visitors (renamed from the old jax backend); delay lines compile to scalar carry entries, `jnp.roll` shift registers, or IOTA-indexed circular buffers depending on size and `-mcd`; `build/backends/jax-only.cmake` builds just these backends. C-semantics math is preserved: `fmod`/`remainder` follow C (truncated / round-to-nearest quotient), not Python floored modulo, and integer-natured outputs are cast to `faust_float`.

Delay-line scalarization: an analysis pass (`NNXScalarizeAnalysis` in `nnx_base_instructions.hh`) finds struct arrays of size `<= 4` whose every access is a single constant index - the `z^-1/z^-2` states of recursive filters - and emits them as individual scalar carry entries with plain assignments instead of `.at[k].set()` dynamic-update-slice writes, which fall off XLA's in-place-update path (a scan body with ten such arrays measured ~1000x slower per step on CPU). Bare whole-array references (the `jnp.roll` strategy) or computed indices (IOTA circular buffers) disqualify an array, so the other delay strategies are untouched, and scalarized stores coerce to the carry entry's dtype so the carry pytree stays stable under `jax_enable_x64`. Measured on CPU (jax 0.10.2, 1 s @ 44.1 kHz): the speedup tracks how much of the per-step work is small recursive arrays. A pure recursive cascade (16 one-pole sections, no large buffers) runs ~1600x faster forward once scalarized; the gain is smaller for DSPs whose per-step cost is dominated by un-scalarizable large delay buffers (e.g. Freeverb, whose comb/allpass buffers stay arrays) and is sensitive to the jax/XLA version's dynamic-update-slice handling.

Architecture files: `architecture/jax/minimal.py` (NNX) and `minimal_linen.py` (Linen) with UI handlers (sliders, nentry via Gumbel-softmax, buttons, checkboxes, soundfiles, hbargraphs and vbargraphs), physical and normalized parameter APIs (`params=` / `normalized_params=`), magic-clamp straight-through gradients, JAX PRNG foreign functions, block-wise `process_block` streaming, polyphony via `jax.vmap`, and safetensors save/load that records the DSP identity (class name plus zone-to-label map) and validates it on load. Documentation in `architecture/jax/README.md` and `DELAY_LINES.md`.

Generated-module API:

- Parameter dicts (`params=` / `normalized_params=`, `with_defaults`, `params_from_labels`) accept zone names, Faust shortnames, or full label paths as keys; unknown keys raise `UnknownParameterError`, partial dicts are merged with the module's defaults, and values are cast to `faust_float` so weakly-typed Python scalars cannot force a jit retrace. Physical `params=` rejects nentry logits/tau entries (they belong to `normalized_params=`) instead of silently ignoring them.
- Widget metadata (`[tau:...], [style:...], [param:...]`) is forwarded to the generated `add_nentry`/`add_soundfile` as an explicit metadata dict, so labels, addresses, and shortnames in `json_metadata` stay clean and shortname-keyed parameter lookup works.
- `process_block` returns `(new_carry, outputs)`, carry-first like tick and Flax [RNN](https://github.com/google/flax/blob/a31129e2f8a2f5ca6ccc8492e536d2b28359f747/flax/nnx/nn/recurrent.py#L47) cells; `__call__` wraps it with a fresh carry. Generators (`num_inputs == 0`) take `num_samples=` instead of an empty input array.
- Class-level introspection without construction: `num_inputs`, `num_outputs`, `json_metadata` (parsed dict), `is_stochastic`, and a generated class docstring listing the parameter table.
- RNG policy: the compiler emits `is_stochastic` (true when the DSP calls `random_*` foreign functions); deterministic DSPs run without any rngs and skip per-sample key threading in the scan, stochastic DSPs raise if no RNG is provided. deterministic (eval-mode nentry argmax vs Gumbel sampling) is a constructor argument. nentry logits supplied through `normalized_params=` are Gumbel-softmax sampled when an nentry key is provided, so their gradients are live.
- `faust_float` coherence: the constructor default tracks the compiled precision (`-single`/`-double`), and state buffers, waveform tables, soundfile buffers, and the initialized carry all follow the runtime `faust_float` dtype (`float32` models stay `float32` under `jax_enable_x64`).
- `[scale:log]` sliders use a geometric normalized mapping whose normalize/unnormalize functions are exact inverses (log scale requires `min > 0`). `[scale:log]/[scale:exp]` nentries carry their scale mode through too.

**Breaking changes for users of the previous `-lang jax` backend:**

- The `-lang jax` flag is removed; use `-lang linen`. `architecture/jax/minimal.py` is now the NNX architecture, so the Linen architecture is `architecture/jax/minimal_linen.py`, and the impulse harness moved to `Make.linen` / `archs/impulse_linen.py`. Old `-lang jax` / `-a minimal.py` invocations must be updated.
- Normalized UI parameters now span [0, 1], not [-1, 1]. Saved parameter values, gradient-optimized param leaves, and `normalized_params=` inputs must be rescaled: a raw normalized value of 0 is the parameter minimum now, where it used to be the midpoint.
- `[scale:log]` and `[scale:exp]` parameters use exact geometric/exponential inverse mappings, so their defaults and normalized<->physical conversions differ from before (the old log mapping initialized, e.g., a log slider with init 1000 on [100, 10000] to ~100).
- `fmod` and integer modulo (%) follow C semantics (the result takes the sign of the dividend, not the divisor), and integer-natured outputs are cast to `faust_float` instead of staying int. DSPs using these on signals that can go negative now produce different output that matches the C++ reference. (remainder is unchanged: both backends use the same round-to-nearest-quotient implementation.)
- The generated module is now carry-based streaming: the old `__call__(x, T)` is replaced by `initialize_carry()` plus `process_block(carry, inputs, ...) -> (new_carry, outputs)` (carry-first), with `__call__(inputs, params=` / `normalized_params=, num_samples=, rngs=)` wrapping a fresh carry. Code calling the old generated module must be rewritten.

Two shared files are touched; both are backward-compatible for every other backend:

- `architecture/faust/gui/JSONUI.h`: `addGenericSoundfile` now flushes pending widget metadata into a "meta" array via `addMeta()`, the way every other widget already does (the soundfile writer was the only one that skipped it, so declared soundfile metadata leaked onto the following widget). The emitted JSON is byte-identical unless a soundfile metadata declare precedes the widget, which the compiler produces only for `-lang nnx`/`linen`; `-json` and C++ runtime JSON for all other backends are unchanged (verified against the cpp backend).
- `compiler/generator/typing_instructions.hh`: the shared `TypingVisitor` gains a `BinopInst` case for whole-array (real-pointer) operands, which the NNX/Linen array-value codegen produces when an unindexed input channel feeds a binop. It sits immediately before the pre-existing "should never happen" assert, so it only assigns a type where the visitor previously aborted, and cannot change the type inferred for any other backend.

Tests and CI: Passing [here](https://github.com/DBraun/faust/actions/runs/29697909705). tests/jax-tests pytest suite (143 tests, uv-managed environment, `.github/workflows/jax-tests.yml`), NNX and Linen impulse-response harnesses (`tests/impulse-tests/Make.nnx`, `Make.linen`) whose second sub-run processes 64-sample blocks to verify block-size independence of the carry state, and compile-tests coverage for both backends. Impulse responses match the C++ reference for both backends.

The Linen impulse-test harness (`tests/impulse-tests/Make.linen`, `archs/impulse_linen.py`) is carried over from the old `-lang jax` harness (`Make.jax`, `archs/impulsejax.py`).


compiler/generator/instructions_compiler.cpp
```diff
@@ -545,7 +546,7 @@ void InstructionsCompiler::compileMultiSignal(Tree L)
             // special handling Julia backend
             pushComputeBlockMethod(IB::genDeclareBufferIterators(
                 "input", "inputs", fContainer->inputs(), ptr_type, false));
-        } else if (gGlobal->gOutputLang != "jax") {
+        } else if (gGlobal->gOutputLang != "nnx" && gGlobal->gOutputLang != "linen") {
```
sletz: Any reason not to use `isPythonBackend` here ?

compiler/generator/instructions_compiler.cpp
```diff
@@ -577,7 +578,7 @@ void InstructionsCompiler::compileMultiSignal(Tree L)
             // special handling for Julia backend
             pushComputeBlockMethod(IB::genDeclareBufferIterators(
                 "output", "outputs", fContainer->outputs(), ptr_type, true));
-        } else if (gGlobal->gOutputLang != "jax") {
+        } else if (gGlobal->gOutputLang != "nnx" && gGlobal->gOutputLang != "linen") {
```
sletz: Any reason not to use `isPythonBackend` here ?

compiler/generator/instructions_compiler.cpp
DBraun: Thanks. Fixed to `!gGlobal->isPythonBackend()`.

compiler/generator/instructions_compiler.cpp
DBraun: Thanks. Fixed to `!gGlobal->isPythonBackend()`.

compiler/generator/instructions_compiler.cpp
sletz: Thanks, I guess this is one of the a last time a human sees something an IA did not ((-;

sletz:
Thanks. Fine for me, anything left before the merge ?

DBraun:
All ready once the actions pass. Squash and merge at your leisure. Thanks!

PR #1279, opened 2026-07-31
wasmtime backend: embeddable build, cheaper math imports, GCC/MSVC fixes

lucaromagnoli:
Three changes the wasmtime backend needs to work as an embedded runtime JIT, found while wiring it up as MAGDA's default Faust backend on macOS, Linux and Windows.

## Embeddable build (`build/CMakeLists.txt`)

`INCLUDE_WASMTIME` hardcoded `/usr/local/lib/libwasmtime.a`, which assumes a system-wide install. `WASMTIME_LIB` and `WASMTIME_INCLUDE_DIR` now let an embedder point at the official prebuilt wasmtime C API artifacts instead. Left unset, the historical `/usr/local` layout is used, so existing builds are unaffected.

## Cheaper math imports

The math functions the WASM backend emits are host imports, so a DSP with transcendentals in its inner loop pays one wasm to host call per sample per call site. Those were registered through `wasmtime_linker_define_func`, whose checked calling convention boxes every argument into a `wasmtime_val_t`. Registering them with `wasmtime_linker_define_func_unchecked` (raw convention: `wasmtime_val_raw_t`, no per-call kind tagging or type check) is materially cheaper.

Measured on macOS arm64 (libfaust 2.85.5, wasmtime v47.0.2, 512-sample buffer) over 44 Faust effects:

- mean speedup over the interpreter backend goes from 4.22x to 9.32x faster
- no effect left slower than the interpreter (reverb and compressor previously sat at 0.85-0.87x)
- output bit-identical to the interpreter on all 44
- worst case for host-call density, a 5-shape antialiased clipper at 10 host calls per sample: 29 to 142 MBytes/sec
- no measurable change on DSP without transcendentals

`wasmtime_val_raw_t` fields are documented as little-endian, so the raw path would need a byte swap on a big-endian host. The checked path is retained in full behind `FAUST_WASMTIME_RAW_MATH=0`.

## GCC and MSVC build fixes

Neither problem is reachable until `INCLUDE_WASMTIME` is on, so both survived until this backend was built on Linux and Windows.

- `std::sinf` and the other `std::<fn>f` overloads are not standard C++. libc++ and MSVC provide them, libstdc++ does not, so GCC rejected all 20 f32 math callbacks. Uses the global `::sinf` forms, with `<math.h>` included for them.
- Three `wasmtime_val_t` designated initializers require C++20. GCC and Clang accept them as an extension, MSVC does not, and this target builds as C++17. Written as plain field assignments instead.

`wasm_dsp_aux.cpp` now compiles clean under `-std=c++17 -pedantic-errors`.

## Testing

Built and run on macOS arm64 and x86_64, Linux x86_64 and Windows x86_64. Bit-exactness checked against the interpreter backend on all 44 effects.

Happy to split this into separate PRs (build plumbing / performance / portability) if that is easier to review.

### DIFF
--- a/build/CMakeLists.txt
+++ b/build/CMakeLists.txt
@@ -45,8 +45,17 @@ option ( LINK_LLVM_STATIC          "Link LLVM statically" on )
 option ( INCLUDE_LLVM_STATIC_IN_ARCHIVE "Combine static LLVM .a files and libfaust.a in a single .a file" on )
 
 if (INCLUDE_WASMTIME)
-    set(WASMTIME_LIB /usr/local/lib/libwasmtime.a)
-    link_directories(/usr/local/lib)
+    # WASMTIME_LIB / WASMTIME_INCLUDE_DIR let an embedder point at a prebuilt
+    # wasmtime C API (the official release artifacts) instead of a system-wide
+    # install. Unset, the historical /usr/local layout is used.
+    if (NOT WASMTIME_LIB)
+        set(WASMTIME_LIB /usr/local/lib/libwasmtime.a)
+    endif()
+    get_filename_component(WASMTIME_LIB_DIR ${WASMTIME_LIB} DIRECTORY)
+    link_directories(${WASMTIME_LIB_DIR})
+    if (WASMTIME_INCLUDE_DIR)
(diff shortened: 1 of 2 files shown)


sletz:
Thanks.

PR #1284, opened 2026-08-22
poly-dsp.h: Steal the oldest vocie, not the least frequently used one

Timothy-Sikes:
Noticed while holding down five 5 keys on a [nvoices:4] patch, and letting each one go individually -- sometimes the first note played got released, sometimes it was a different one.

Looks like the code was tracking and checking usage instead of fDate being a clock. Instantiation also seems to indicate the variable was for a clock rather than tracking usage, and "steal the oldest voice" seems more reasonable and more in line with the intention rather than the least frequently used voice.

### DIFF
--- a/architecture/faust/dsp/poly-dsp.h
+++ b/architecture/faust/dsp/poly-dsp.h
@@ -624,7 +624,7 @@ class mydsp_poly : public dsp_voice_group, public dsp_poly {
         // Allocate a voice with a given type
         int allocVoice(int voice, int type)
         {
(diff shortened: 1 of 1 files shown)


Timothy-Sikes:
Claude generated repro using the JS wasm version: https://gist.github.com/Timothy-Sikes/e0564b08f99bbf57c732867de2d85a4b

sletz:
Fixed in https://github.com/grame-cncm/faust/commit/cc0acf83ca6ed1bd65fa56bf276d90a7467fa2e4 with a test.

PR #1286, opened 2026-08-26
paradigma.dsp example update  with IR matched noise and other improvements

gmvox:
Updated example s/ misc/ paradigma.dsp
with IR matched synched noise, better code and structure and more comments


gmvox:
Thank You.

PR #1287, opened 2026-08-30
poly-dsp.h: fade the second half of a stolen voice in, not only the first half out

crafics:
Follow-up to grame-cncm/faustwasm#43 / #44, as asked there: the same fix for the C++ architecture the TS port follows.

When `mydsp_poly` steals a voice, `mydsp_poly::compute` renders that voice in two halves (`dsp_voice::computeLegato`): the old note on the first half, which `fadeOut` ramps down to `1/count`, and the new note on the second half from the DSP's live state after `keyOn`. `fadeOut` scales the buffer, not the voice — an envelope that has not reached silence carries on at its level, so the second half started at full level. That is a step on every stolen note; on a monophonic patch, whose one voice is never silent between notes, a click on every note after the first.

**Fix:** `fadeIn(count, offset, buffer)`, the mirror of `fadeOut` (`1/count` up to 1), applied to the second half right after `computeLegato` — one line in `compute`, one new method. The two halves now meet at `1/count` of the running level: the stolen note's attack is ramped over half a block, the way the outgoing note's release already is. While there: for an odd `count` the second half takes the remainder (`computeSlice(slice, count - slice)`), so the last frame is rendered instead of keeping the previous block's value (the TS port already does this).

**Test:** `tests/polyphonic-tests/legato-fade-repro.cpp`, a sibling of `voice-steal-repro.cpp` (standalone against `poly-dsp.h`, no generated code): one voice with a one-pole envelope, note 62 steals note 60, largest sample-to-sample step in the stolen block.

| | samples around the split | largest step |
|---|---|---|
| before | `0.0115 \| 0.738` | 0.727 |
| after | `0.0115 \| 0.0115` | 0.013 (the ramp's own increment) |

`make check` in `tests/polyphonic-tests` runs both reproducers; on an unpatched tree the new one exits with failure, which is the intended proof.

### DIFF
--- a/architecture/faust/dsp/poly-dsp.h
+++ b/architecture/faust/dsp/poly-dsp.h
@@ -225,8 +225,8 @@ struct dsp_voice : public MapUI, public decorator_dsp {
         // Start next keyOn
         keyOn(fNextNote, fNextVel);
         
-        // Compute on second half buffer
-        computeSlice(slice, slice, inputs, outputs);
+        // Compute on second half buffer (the remainder, for an odd count)
+        computeSlice(slice, count - slice, inputs, outputs);
     }
 
     // Extract control paths from fullpath map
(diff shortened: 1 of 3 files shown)


sletz:
Merged, thanks.

PR #1303, opened 2026-09-20
Fix Julia backend double precision cosh symbol

### LINKED ISSUE #1298: Julia backend: compiling DSP with cosh in double-precision mode fails with undefined symbol error
### Problem

When compiling a Faust DSP that uses hyperbolic cosine (`cosh`) targeting Julia in double precision (`-double`), compilation fails with an undefined symbol error:

```
test.dsp:1 : ERROR : undefined symbol : cosh
```

### Root Cause

In `compiler/generator/julia/julia_instructions.hh` (around line 198), under the double-precision registration section of `JuliaInstVisitor`, `coshf` is registered instead of `cosh`:

```cpp
// Double version
...
// Hyperbolic
gFunctionSymbolTable["acosh"] = true;
gFunctionSymbolTable["asinh"] = true;
gFunctionSymbolTable["atanh"] = true;
gFunctionSymbolTable["coshf"] = true;  // Typo: should be "cosh"
gFunctionSymbolTable["sinh"]  = true;
gFunctionSymbolTable["tanh"]  = true;
```

Because `"cosh"` is missing from `gFunctionSymbolTable` (and `"coshf"` was registered twice: once under float and once under double), the compiler treats `cosh` as an unknown function when generating double-precision Julia code.

### Steps to Reproduce

Create a minimal DSP:
```faust
process(x) = cosh(x);
```

Compile with:
```bash
faust -lang julia -double test.dsp
```

Output:
```
test.dsp:1 : ERROR : undefined symbol : cosh
```

### Expected Behavior

The DSP should compile successfully to Julia double-precision code without errors.

### Suggested Fix

In `compiler/generator/julia/julia_instructions.hh`, update line 198:
```cpp
gFunctionSymbolTable["cosh"] = true;
```

h30s:
Fixes #1298

In `compiler/generator/julia/julia_instructions.hh`, `coshf` was mistakenly registered instead of `cosh` in the double-precision function symbol table. This caused Julia backend compilations using `cosh` with `-double` to fail with an undefined symbol error.

This PR fixes the table entry to register `cosh`.


sletz:
Thanks.

